PR #4, opened 2024-04-24
Merge back from the LSIR/llm--mediator-political repository

TiboX2021:
Merge back from the https://github.com/LSIR/llm-mediator-political repository.

Changes since separation:

- Webapp refactor in order to work with cloud functions as a backend and not a local state + removed unused code
- Large refactor for code readability
- Cloud functions integration as a backend
- Firestore integration
- Fully functional Firebase backend deployed @ EPFL

The Webapp transition to backend is not complete.

Refactored & working stages:
- TOS & Profile
- Survey
- Chat about items (real-time)

Not yet refactored:
- Vote for leader
- Vote reveal

Note that the last 2 stages will be easy to refactor, as they are essentially simple forms.


LeoLaugier:
Ready to merge if it looks fine for you @cjqian

cjqian:
Thanks for the clear documentation of the changes! LGTM.

EPFL.md
```diff
@@ -0,0 +1,38 @@
...
+```bash
+firebase deploy  # Deploys everything (Cloud Functions, Firestore rules)
+firebase deploy --only functions  # Deploys only Cloud Functions
+```
+
+You can then access various consoles at the following URLs:
+
+| Console                 | URL                                                                                                       |
+| ----------------------- | --------------------------------------------------------------------------------------------------------- |
+| Project Console         | https://console.firebase.google.com/u/0/project/friendlychat-d6dc5/overview                               |
```
iislucas, line 20: Lets try and factor our project IDs from documentation also. (i.e. there should be no mention of friendlychat-d6dc5 in the repo)

README.md
```diff
@@ -1,3 +1,151 @@
...
+
+- Authentication UI: `http://localhost:4000/auth`
+- Emulator UI: `http://localhost:4000`
+- Firestore UI: `http://localhost:4000/firestore`
+- Cloud Functions UI: `http://localhost:4000/functions`
+
+### Firestore
+
+We use Firestore as our database. The rules are located in the [`firestore.rules`](./firestore.rules) file.
+Basically, the database cannot be externally accessed, and must be interacted with through cloud functions.
```
iislucas, line 75: This is one way to do it, but often you can significantly simplify your cloud functions if you have a smart set of rules.

README.md
```diff
@@ -1,3 +1,151 @@
...
+
+A database prototype schema can be found [here on dbdiagrams.io](https://dbdiagram.io/d/Firebase-LLM-Mediation-660d473a03593b6b61123f24) (readonly, change it with your own if you take-on the project).
+
+You can seed the database with the default data by running the following command:
+
+```bash
+curl "http://127.0.0.1:5001/friendlychat-d6dc5/us-central1/seedDatabase?seeder_password=seeder_password"
```
iislucas, line 82: I would prefer to have a script that we run locally that talks to the DB for initialization, rather than a web-hook with a magic password. Generally we should lean on existing authentication protocols, instead of making our own with passwords. You can do this quite easily because the cloud sdk authenticates with the project, so you can have a locally run script that will do initalization and remove the webhook for seedDatabase.

README.md
```diff
@@ -1,3 +1,151 @@
...
+
+Upon running `npm run build`, if the emulator is running, it will automatically reload the functions.
+
+### Authentication
+
+This project sets up Firebase Authentication with email/password and Google sign-in.
+
+#### Email - Password authentication
+
+By default, experiment participants log in using their `uid`. It is passed to the Firebase Authentication service as `email:{uid}@palabrate` and `password:{uid}`.
```
iislucas, line 106: I think we'd be better to avoid participants having to have a user account on firebase. If they have that, then we need to do a lot more user-management work. Users being completely anonymous will be much better for us.

iislucas: (so then we'd say that: only admins/experimenters have an account on firebase)

README.md
```diff
@@ -1,3 +1,151 @@
...
+
+#### Google sign-in
+
+The Google sign-in is used for the administrators. They can log in using their Google account.
+
+In order to create a dummy google account when running the project locally with emulators, follow these steps:
+
+1. Launch the frontend and click on the Google sign-in button.
```
iislucas, line 114: I would recommend we make a TODO to make this into a small typescript script that does this for us so we don't have to depend on UI, and we can later better automate local testing/management

firestore/firestore.rules
```diff
@@ -0,0 +1,27 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{database}/documents {
+
+    // Rules for the messages collection
+    match /messages/{messageId} {
```
iislucas, line 6: shouldn't this be a bit more nuanced.... e.g. /messages/{experimentID}/groupChat/{groupChatId}/{messageId} or something like that?

functions/.env.example
```diff
@@ -0,0 +1 @@
+SEEDER_PASSWORD=seeder_password
```
iislucas, line 1: Lets try and remove this file / need for a seeder password.

iislucas (commented):
A few comments to see if we can improve the design some / future proof it for when we want real users/experiments.

iislucas:
@cjqian see comments here for roadmap planning.

README.md
LeoLaugier: Currently, a participant only sees the uid they are assigned, after the experimenter generates the uids and shares with the usparticipant. So technically, users never interact with their accounts besides logging in with their uid (the dummy password is handled by the frontend app). 

Otherwise, we may have trouble if users accidentally close their browser for a short period of time (by the way we should discuss the expected behavior when users abandon an experiment).

iislucas: We should check with Crystal, but I think they get sent their UID by prolific/the expérimenter, so we should not need to link their email to it on firebase... 

I think the key security consideration and assumption here is that participant uniqueness is defined by unique URL (basically we use a security model for participants like photos links). 

This is my best estimate of the right balance on balancing pragmatics, privacy, and security. I'll also make sure to do a review with security/privacy experts too.

EPFL.md
TiboX2021: done in #8 - 6c9d812932eade4f8ccee8fca2dd83da0f103edb

firestore/firestore.rules
TiboX2021: You are right, I defined the current firestore layout by lazily translating the previous draft `AppState` with my limited knowledge of firestore which made it too monolithic in some places and overly nested in others. Now that the app completely relies on Firebase and that I understand it better, I will take some time to think about a better database structure and take this into account.

README.md
TiboX2021: I will think of a better firestore database layout. I can probably replace many (if not all) of the "fetcher" cloud functions by clever firestore rules and frontend logic. However, would you recommend doing this for mutating cloud functions ? I guess that in this particular case, it would be better to keep the business logic in cloud functions that can apply data validation in a secure way.

iislucas: It depends... ideally the server doesn't have to be updated much, and the server-side functions can be quite generic. I'll look over in more detail later. Sometimes you can get away with no server-side logic beyond some security rules, sometimes you need some server-side functions. @cjqian FYI

README.md
TiboX2021: Fixed in #8 - 94e43b38773459f1d1e078ff98a42155c22840d6
Participants will "log in" using their UID, without Firebase Authentication

README.md
TiboX2021: I have not found a way to create a fake user with Google.com as an authentication provider programmatically yet, I created an issue for this: #9

PR #8, opened 2024-04-25
Implement vote and reveal & other improvements

### LINKED ISSUE #7: Hide API key in a local file
It shouldn't be an issue to make Firestore API key public since we have an authentification system but for peace of mind, we should hide the [API Key](https://github.com/PAIR-code/llm-mediation-experiments/blob/3e06de8a8d8a2841e1a036b074cf72a664abf539/webapp/src/lib/api/firebase.ts#L11) in a local file.

### LINKED ISSUE #9: Add script to generate fake experimenters with google.com as a provider
Experimenters authenticate through the google.com popup (which also allows other auth providers). We should create a script to generate fake experimenters with auth providers in order to facilitate development and testing.

However, it seems that creating fake users with an auth provider programmatically is not easy (I did not find how to do it).

TiboX2021:
- [x] Finished the implementation of stages of type `voteForLeader` and `leaderReveal` in order to finish the transition from local demo to backend-based webapp.

- [x] Updated the docker config for the webapp (documentation included in README.md)

- [x] Removed the hardcoded firebase configurations for deployment @ EPFL and included example configuration files (documented in README). Resolves #7 

- [x] Take into account remarks made in #4's thread 

Commit c8384745c1d0cc2b075dfd9347e4f16a3f80b544 resolves #9 

Firestore & cloud functions refactoring will be done in another PR


.vscode/settings.json
```diff
@@ -23,7 +23,7 @@
       "splitTerminals": [
         {
           "name": "firebase",
-          "commands": ["firebase emulators:start"],
+          "commands": ["firebase emulators:start --import ./export"],
```
iislucas, line 26: lets name the export file something more meaningful. e.g. "emulator_test_config" (assuming that's what is in it, which I think it is...).

README.md
```diff
@@ -20,6 +18,8 @@ This is a repository to support collaboration on using LLMs in behavioral econom
 │
 ├── docs       # Documentation
 │
+├── export     # Firebase Authentication export with default google accounts for Auth emulator
```
iislucas, line 21: fix name here too.

README.md
```diff
@@ -50,16 +52,26 @@ Install the firebase cli tools with the following commands:
...
+### Configuration
+
+Create the configuration files for a default firebase project:
+
+```bash
+cp .firebaserc.example .firebaserc
+cp webapp/src/lib/api/firebase-config.example.ts webapp/src/lib/api/firebase-config.ts
 ```
 
+This should be enough for local development with emulators. Be sure to input the correct project id and keys once you actually deploy the project to firebase!
```
iislucas, line 67: Clarify a little more: be sure to update the project ID in the .firebaserc file and keys in ....

README.md
```diff
@@ -50,16 +52,26 @@ Install the firebase cli tools with the following commands:
...
+
 ### Emulators
 
 In order to run offline and for development purposes, we use Java Firebase emulators.
 
 ```bash
-export JAVA_TOOL_OPTIONS="-Xmx4g"  # Set your desired max RAM (here: 4GB)
-firebase emulators:start  # Start the emulators
+firebase emulators:start --import ./export  # Start the emulators and load the default Auth dump
```
iislucas, line 74: fix export filename here too.

README.md
```diff
@@ -138,14 +145,30 @@ Run `ng generate component component-name` to generate a new component. You can
...
+Build the webapp with the following command:
 
-### Further help
+```bash
+# Default build / test the image locally
+ng build --configuration=production
 
-To get more help on the Angular CLI use `ng help` or go check out the [Angular CLI Overview and Command Reference](https://angular.io/cli) page. This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 16.2.7; it was then updated to Angular 17.
+# Serve the app from the /myapp prefix if you use a k8s cluster that shares the same domain for multiple services.
+ng build --configuration=production --base-href /myapp/
```
iislucas, line 161: Presumably you can run it locally with `ng serve` still also? I think that instruction is probably helpful for fast iteration on the UI (e.g. it has live reload); I think docker based building and serving is useful for easy generic deployment; but probably not needed fore firebase... ?

export/auth_export/accounts.json
```diff
@@ -0,0 +1,47 @@
...
+        {
+          "providerId": "google.com",
+          "rawId": "1651642867512372825802162318175962691422",
+          "federatedId": "1651642867512372825802162318175962691422",
+          "displayName": "Experimenter",
+          "email": "experimenter@google.com"
+        }
+      ],
+      "lastRefreshAt": "2024-04-29T14:23:03.451Z",
+      "salt": "fakeSalttpRdKDA4NS5sxXb2GeoT",
```
iislucas, line 21: Can these fields be skipped?

iislucas: (sald and passwordHash), I'd generally rather avoid such info in a repo, even if it's fake. Or is this needed to login to the emulator?

scripts/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "scripts",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "seed-db": "npx ts-node src/seed-database.ts",
```
iislucas, line 6: is seed-db still used?

scripts/src/admin.ts
```diff
@@ -0,0 +1,35 @@
+/** Load the Firebase app and expose it to scripts */
```
iislucas, line 1: maybe note in the comment the assumption about where this is being run. (I guess it assumes it's being executed from the scripts directory?

scripts/src/seed-database.ts
```diff
@@ -0,0 +1,367 @@
...
+  console.log("Done !");
+};
+
+// ********************************************************************************************* //
+//                                         SEEDER DATA                                           //
+// ********************************************************************************************* //
+
+// Note that this data is hardcoded for now, because the db structure will change in order to be more practical.
+
+const DEFAULT_STAGES = [
```
iislucas, line 65: It's probably good to have this somewhere general, so it can be used in tests, and here, and also to make sure the constant is type-checked; e.g. 

const DEFAULT_STAGES: <the type of a stage>[] = [ ...

scripts/src/seed-database.ts
```diff
@@ -0,0 +1,367 @@
...
+        },
+      ],
+    },
+  },
+];
+
+const DEFAULT_STAGE_MAP = DEFAULT_STAGES.reduce((acc, stage) => {
+  acc[stage.name] = stage;
+  return acc;
+}, {} as Record<string, any>);
```
iislucas, line 247: Don't use `any` types - typescript has great typing support, aim to use it as much as you can. :) 

You can use `Object`, or `unknown`. Better to use "unknown" and then later to typecast to what you want it to be explitily, e.g. `... as Stage[]`. 

Also make sure to add the type for `DEFAULT_STAGE_MAP` too.

scripts/src/seed-database.ts
```diff
@@ -0,0 +1,367 @@
...
+    "4. Post-chat survey": true,
+    "5. Post-discussion leadership survey": true,
+    "6. Vote for the leader": true,
+    "7. Post-discussion work": true,
+    "8. Leader reveal": false,
+    "9. final satisfaction survey": true,
+  },
+};
+
+const DEFAULT_PARTICIPANTS = [
```
iislucas, line 265: Make sure this is typed also.

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
```diff
@@ -72,12 +73,51 @@ const TIMER_SECONDS = 60; // 1 minute between item pairs for discussions
...
+
+    // Initialize the current rating to discuss with the first available pair
+    const { id1, id2 } = this.stage.config.ratingsToDiscuss[0];
+    this.currentRatingsToDiscuss = signal({
+      item1: this.stage.config.items[id1],
+      item2: this.stage.config.items[id2],
+    });
+
+    this.unsubscribeMessages = chatMessagesSubscription(this.stage.config.chatId, (m) => {
+      this.messages.set(extendUntilMatch(this.messages(), m.reverse(), 'uid'));
```
iislucas, line 91: We should maybe chat about using `extendUntilMatch`; from looking at the code, I think it's not strictly an extension; it seems to take old elements up to the when the 'key' (in this case 'uid' in new list matched the old list, and then concat from that point on)... this makes me think this logic is probably slightly wrong as I think two messages by the same person will result in the cut being too early... but I'm not sure if I'm getting the logic... 

I guess my expectation here is that new messages just replace the old ones... why do we need/benefit from this fancy concat logic?

Maybe good to comment on what "m" is, is it a single message? or a set of messages? Maybe a longer variable name would help?

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
```diff
@@ -72,12 +73,51 @@ const TIMER_SECONDS = 60; // 1 minute between item pairs for discussions
...
+
+      // Find if new discuss items message have arrived
+      const last = m.find((m) => m.messageType === MessageType.DiscussItemsMessage) as
```
iislucas, line 94: Don't use the name name for a list of things as a thing in the list :)  (m in this case).  Use a bit more detailed variable names and it'll be easier to understand.

iislucas: I'm also worried that this will simply find the last pair of items, not tell you if there's a new pair of items to discuss...

webapp/src/app/participant-view/participant-stage-view/exp-leader-reveal/exp-leader-reveal.component.ts
```diff
@@ -1,80 +1,66 @@
...
-    //     .every((n) => n === 1);
-    //   // Allow "Next" to be pushed.
-    //   if (isReady) {
-    //     for (const user of users) {
-    //       user.allowedStageProgressionMap[user.workingOnStageName] = true;
-    //     }
-    //   }
-    //   return isReady;
-    // });
+    this.everyoneReachedThisStage = signal<boolean>(false);
```
iislucas, line 52: Maybe add a comment to say that this is signal is expected to be thrown away and replaced OnInit when the stage data is set... ?

iislucas: Might be better to simply not have this initialized... otherwise you risk creating a dependency on the signal and then having the signal value replaced, but the dependency update forgotten about... Alternatively, whenever the stage is set, you write updates from the stage's signal to this signal... (that might be better, then you have the value always defined)

webapp/src/app/participant-view/participant-stage-view/exp-leader-reveal/exp-leader-reveal.component.ts
```diff
@@ -1,80 +1,66 @@
...
+  // Reload the internal logic dynamically when the stage changes
+  @Input({ required: true })
+  set stage(value: ExpStageVoteReveal) {
+    this._stage = value;
+
+    this.stageData = this.stage.config;
+    this.everyoneReachedThisStage = this.participant.everyoneReachedCurrentStage(this.stage.name);
+  }
+
+  private queryClient = injectQueryClient();
```
iislucas, line 29: Needs a comment to say what this is I think.

iislucas: ... I suspect this is not the right design... I think you should be simply be doing a firebase groupQuery for the participants, and then when they are all at done at this stage, you move to the next...

webapp/src/app/participant-view/participant-stage-view/exp-leader-reveal/exp-leader-reveal.component.ts
```diff
@@ -1,80 +1,66 @@
...
-    //       ['vote'],
-    //     ),
-    //   );
+    this.finalLeader = signal<string>('TODO');
+  }
 
-    //   return sorted[0].userId;
-    // });
+  nextStep() {
+    this.mutationReveal.mutate({
```
iislucas, line 59: I think we need a comment to say what/why we are doing this...

iislucas (commented):
Great to see the progress! 

I think there's a few patterns I don't quite get (e.g. the mutation at end of stages when within the nextStep()), might need a video chat about that. 

Some comments/suggestions in the meantime...

.vscode/settings.json
TiboX2021: fixed in 9a95290e5d04a9f824d3501641fb11e7b25e52ff

README.md
TiboX2021: fixed in 9a95290e5d04a9f824d3501641fb11e7b25e52ff

README.md
TiboX2021: Yes, these actually are the production deployment instructions, I forgot to mention the usual `ng serve` / `npm run start` for development...

TiboX2021: clarified in 14e7118ccb970d41a3fd2a439e2538d472e289a4

export/auth_export/accounts.json
TiboX2021: fixed in f5a7d3eb062455978049c0c7aee1e253cddc5cf7

scripts/package.json
TiboX2021: yes, it populates firestore with an initial template + experiment config

scripts/src/admin.ts
TiboX2021: clarified in 6c546282352ff23843530efb9480eddaff2accda

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
TiboX2021: currently, when a participant opens a chat stage, they subscribe to the `n` most recent messages from firestore. On every update, the frontend appends new messages to its list (because from my understanding of firestore, the situation where 2 messages being created in a short span could only trigger one subscription event can happen ?). We can indeed discuss of better ways to handle incoming messages

TiboX2021: `extendUntilMatch` assumes that 2 arrays of objects are "sorted" in a compatible order, and takes from the first array until an object with the same key value as the second array's first element is reached. It then appends all of the second array. It does not do so "in-place", so I agree that "extend" may not be the right word for it...

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
TiboX2021: renamed in 8ef5c53cde2ec9a8f4761e68748efa3b174b8f00

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
iislucas: But isn't "uid" a user's ID? or is it a unique firestore object ID? I think if you did this with a unique message ID, it might be right... But I would suggest a simpler way to do this: take the messages, push into a map by message ID, and then Object.assign to that to get the union, and then re-flatten by message sent timestamp? Then you don't depend on subtle semantics of extendUntilMatch... An important part of this kind of coding is making the semantics as easy to interpret as possible (more time is generally spent reading code than writing code, so optimise for each of reading) ) :)

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
TiboX2021: I track the current pair being discussed through the `participants_ready_to_end_chat` sync document, which also stores the index of the current pair being discussed in the preconfigured chat list. I should remove one of these 2 ways of computing the last pair being discussed, as they are redundant... If we want to keep the one that scans the last `DIscussItemsMessage` as the single source of truth, we could simply use a `WritableSignal` with lodash's deepEq comparison to make sure that the signal triggers on actual changes

webapp/src/app/participant-view/participant-stage-view/exp-chat/exp-chat.component.ts
TiboX2021: in this code, `uid` stands for `unique ID`, I agree that it is misleading. Perhaps using `mid` for `message ID` would be better ?

iislucas: Ah, we may not control that if it's a firebase value... but at least a comment to clarify since uid is so often used for user-id also.

TiboX2021: I could define this function and use it instead of `extendUntilMatch`
```typescript
export const mergeByKey = <T extends object, K extends Indexable<T>>(
  previousArray: T[],
  incomingArray: T[],
  key: K,
): T[] => {
  return Object.values({ ...lookupTable(previousArray, key), ...lookupTable(incomingArray, key) });
};
```

TiboX2021: done in fbd6f6420695b44e45e2662ed8927502f859e681

webapp/src/app/participant-view/participant-stage-view/exp-leader-reveal/exp-leader-reveal.component.ts
TiboX2021: yeah, the idea was to structure the stage flow like so:
1. You fill the stage data as a participant
2. You use a mutation to send your data to the backend / firestore
3. If the mutation was successful, you navigate to the next page

However, I realized later on that participants are not supposed to be able to edit previous stages, so we should be able to simplify this (if the stage is being worked on by the participant, mutation -> navigation. Else, just navigate).

Because the LeaderReveal stage has no data to be filled by the participant, this mutation feels especially heavy (all that work just to update `workingOnStageName`), I will think of lighter alternatives

iislucas: oh I think I see, the point here is that you save the data to firestore at the time the person does the nextStep thing. Previously I did writes when data was edited... might be worth discussing how hard this is, and what the right abstraction is. My default would be that each stage kind has a class, and that class has functions to edit the data, and when edited, that results in a write to Firebase. So the key thing in a nextStep would be to call the corresponding stage instance's nextStep method that would simply mark this stage as complete.

iislucas (approved):

cjqian:
Wow, thank you!

PR #58, opened 2024-06-18
Add LLM service (Gemini API) to frontend

### LINKED ISSUE #51: Add Gemini API service to Lit frontend

vivtsai:
This sets up ChatService to query the Gemini API for an LLM mediator response after each new chat message and (if no other chat messages have been sent in the meantime) converts that response to a mediator chat message.

> Note: This uses a placeholder prompt, which we'll have to update for real mediation :)

image

Fixes #51

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -33,10 +33,10 @@ cd lit
 npm install
 ```
 
-Then, set up a Firebase config (fork the example and fill in details):
+Then, set up Firebase and LLM config (fork the example and fill in details):
 
 ```bash
-cp src/shared/firebase_config_example.ts src/shared/firebase_config.ts
+cp src/shared/config_example.ts src/shared/config.ts
 ```
 
 Finally, run the app (with live reload):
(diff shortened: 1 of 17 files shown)


TiboX2021:
I think I found a small bug in the way survey stages compute whether they have been fully completed.
In [this line](https://github.com/vivtsai/llm-mediation-experiments/blob/6845fb5960d072eff682dcc62309d46f64e9f451/lit/src/experiment-components/survey/survey_preview.ts#L51), shouldn't the `map` be a `filter` as well ?

vivtsai:
> I think I found a small bug in the way survey stages compute whether they have been fully completed. In [this line](https://github.com/vivtsai/llm-mediation-experiments/blob/6845fb5960d072eff682dcc62309d46f64e9f451/lit/src/experiment-components/survey/survey_preview.ts#L51), shouldn't the `map` be a `filter` as well ?

Yes, good point! Let me push a fix for it

PR #62, opened 2024-06-18
Fix small access issues on frontend

vivtsai:
- [x] Enable experiment deletion from experiment page (not just Home page)
- [x] Allow all users to write mediator chat messages (since right now they are sent from the frontend)
- [x] Gate "leader election" component with "progress-stage-waiting" (wait for all participants to get to this stage, since they must have their profiles set)


cjqian (approved):
Looks good!

PR #63, opened 2024-06-20
Docker deployment for the lit webapp

### LINKED ISSUE #36: Fix hosting local item assets
As a temporary workaround, host online.

TiboX2021:
Fixes #36

Docker setup for serving the lit app.

Main issue: make it possible to add a prefix to all relative urls in the app for deployment to a domain name with prefix (e.g https://modemos.epfl.ch/palabrate instead https://modemos.epfl.ch/ )


lit/README.md
cjqian, line 1: Do you mean URL prefix like "https://modemos.epfl.ch/"? Looking at the code above for image hosting, have you tested locally what happens if this is not set (e.g. do images point to the right relative path if the URL is "" + image/path.jpg)?

cjqian (approved):
Looks great, thank you! Just one question on the URL prefixes - it'd be nice to add an example on the README for it as well.

lit/README.md
LeoLaugier: I'm understanding the prefix to be just `palabrate` in our case:
`URL_PREFIX=/palabrate/ npm run build`

@TiboX2021 I'll let you correct me if I'm wrong and also answer if you have tested locally with no prefix

TiboX2021: No, it has to be surrounded with "/" due to the way html parses <base/> urls. I have tested it.

TiboX2021: @cjqian documentation is in `lit/README.md`

PR #79, opened 2024-06-26
Add LLM mediator config options to chat stage

### LINKED ISSUE #12: Add LLM settings to chat stage config
- [ ] Create and save LLM Mediator though its options
- Mediator ID
- model_name (as defined by the API, e.g. "text-bison@001")
- LLM API service
- prefix
- suffix
- [ ] scrolling menu to select existing Mediator IDs

vivtsai:
Fixes #12

### DIFF
--- a/lit/src/services/firebase_service.ts
+++ b/lit/src/services/firebase_service.ts
@@ -23,7 +23,7 @@ import {
     FIREBASE_LOCAL_HOST_PORT_FIRESTORE,
     FIREBASE_LOCAL_HOST_PORT_FUNCTIONS
 } from '../shared/constants';
-import { FIREBASE_CONFIG } from '../shared/config';
+import { FIREBASE_CONFIG } from '../shared/config_example';
 
 import { Service } from "./service";
 
--- a/functions/src/triggers/stages.triggers.ts
+++ b/functions/src/triggers/stages.triggers.ts
@@ -32,7 +32,7 @@ export const initializePublicStageData = onDocumentWritten(
         break;
       case StageKind.GroupChat:
         // Initialize the custom chat data (depending on the chat kind)
-        let chatData: PublicChatData | undefined;
+        let chatData: PublicChatData;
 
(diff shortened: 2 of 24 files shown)


LeoLaugier:
@vivtsai could you confirm that you have no issue when interacting with the database when seeding it or saving a new experiment?

lit/src/services/firebase_service.ts
```diff
@@ -23,7 +23,7 @@ import {
     FIREBASE_LOCAL_HOST_PORT_FIRESTORE,
     FIREBASE_LOCAL_HOST_PORT_FUNCTIONS
 } from '../shared/constants';
-import { FIREBASE_CONFIG } from '../shared/config';
+import { FIREBASE_CONFIG } from '../shared/config_example';
```
GnRlLeclerc, line 26: You should not import the example config, but copy it to the example file (as indicated in the lit readme). Léo and I are working with the project deployed at EPFL and we need the code to be importing our own config.

cjqian: +1

cjqian:
Looks good to me to merge, pending confirmation that this doesn't break experiment creation. Please test on local emulator and the live Firestore!

lit/src/services/firebase_service.ts
vivtsai: oh yes, whoops! an accidental add, let me revert this one. thank you for catching

vivtsai:
Okay! Confirmed that this works with seed-database and locally. Do we want to deploy the updated functions to Firebase so that it will work with the real Firestore? @cjqian

cjqian:
> Okay! Confirmed that this works with seed-database and locally. Do we want to deploy the updated functions to Firebase so that it will work with the real Firestore? @cjqian

Sounds good! Feel free to do this now, or we can do this tomorrow morning to sync all changes.

LeoLaugier:
Thank you!

PR #80, opened 2024-06-26
Fix chat side panel styling

### LINKED ISSUE #75: Fix chat side panel styling
In chat left side panel,

- [ ] Change panel max-width to width (so that it doesn't shrink when the discussion disappears)
- [ ] Add divider between item discussion preview and "ready to end chat" button (similar to the one between participants and discussion)

GnRlLeclerc:
In chat left side panel,

- [ ] Change panel width (so that it doesn't shrink when the discussion disappears) --> add `flex-shrink: 0` to `.panel` class
- [x] Add divider between item discussion preview and "ready to end chat" button (similar to the one between participants and discussion)

Fixes #75

### DIFF
--- a/lit/package-lock.json
+++ b/lit/package-lock.json
@@ -39,6 +39,7 @@
         "process": "^0.11.10",
         "sass": "^1.74.1",
         "sass-loader": "^14.1.1",
+        "ts-lit-plugin": "^2.0.2",
         "ts-loader": "^9.5.1",
         "typescript": "^5.4.4",
         "webpack": "^5.91.0",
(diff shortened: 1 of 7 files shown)


vivtsai:
Any chance you could also fix the panel width? Adding `flex-shrink: 0` to `.panel` will do it.

https://github.com/PAIR-code/llm-mediation-experiments/blob/593bef2b1428ffd18b2179815cab0a1c97b086ae/lit/src/experiment-components/chat/lost_at_sea_chat.scss#L15

vivtsai:
> Any chance you could also fix the panel width? Adding `flex-shrink: 0` to `.panel` will do it.
> 
> https://github.com/PAIR-code/llm-mediation-experiments/blob/593bef2b1428ffd18b2179815cab0a1c97b086ae/lit/src/experiment-components/chat/lost_at_sea_chat.scss#L15

Actually, just kidding: I'm touching this file in my current commit, so I can do it (if that helps).

Issue #17, opened 2024-05-02
Add implicit reveal stage (to show results, solutions, and team score)


cjqian:
In the vote reveal panel, display

1. Who was elected as the team leader
2. The final resulting task order
3. The correct task order (with an explanation or info icon popup?)
4. A scoring indicator (see appendix o

![Image](https://github.com/PAIR-code/llm-mediation-experiments/assets/4844633/5632a1c1-5f37-47ed-9b3c-14d9899de4c8)

f https://docs.google.com/document/d/11IJXf-tZVKZp592-MNM3hIj9Rft6cBYV7UcYFS9VJJc/)

vivtsai:
Updated proposal (see components doc) for this implementation:

1. Add result/scoring configuration settings to relevant existing stages (e.g., leader election stage)
2. Add an *implicit stage* for displaying all reveals/results (automatically added to experiment if 1+ stages have results/scoring configured)

vivtsai:
The Lit frontend has an MVP "reveal" stage, i.e., results preview for election (vote for leader) stage:

`lit/src/experiment-components/modules/election/election_reveal.ts`

Ideally, we want this election preview component, along with new components for Lost at Sea reveals, to be rendered inside a new implicit Reveal stage component.

Note that setting this up will involve:

(a) **adding implicit-ness**: cleaning up experiment_config and utils logic around adding the reveal stage (e.g., make sure exactly 1 reveal stage is present iff. at least one of the other stages has results configured)

(b) **showing multiple reveals**: updating (or replacing) the current RevealVoted stage in backend (which can only be tied to one VoteForLeader stage - instead, we'll want it to link to all stages with results). Actually, the fastest way to do this might be to ignore this backend field and instead select stages on the frontend (i.e., when rendering the new Reveal component, just check all the other stages to see which ones have results configured)?

cjqian:
Updated design here: https://docs.google.com/document/d/1mQAhpm7rW1xWUhYFcjWHlKdZmOe12HWc0OZSeK6UIlk/edit?tab=t.0

PR #137, opened 2024-07-16
Add composite reveal stage (to show election and LAS ranking results)

### LINKED ISSUE #135: Use unique IDs for experiment stages
Instead of relying on numbering in stage names, have two separate fields: unique IDs and (editable) stage names. Add a field to specify ordering via stage IDs.

vivtsai:
Fixes #135 #17

Changes include:
- [x] Define experiment stages by unique ID instead of name (this removes reliance on numbering at the front of stages)
- [x] Add stage fields to specify whether stage is "composite" (reliant on data from other stages) or not and if it is associated with a game ID  
- [x] Allow users to configure Reveal stages by specifying and ordering other stages (e.g., election, LAS survey) to show results for


utils/src/types/stages.types.ts
```diff
@@ -27,14 +26,17 @@ export const ALLOWED_STAGE_PROGRESSION = {
...
   [StageKind.TakeSurvey]: true,
-  // [StageKind.RankItems]: false,
 } as const;
 
+
 export interface BaseStageConfig {
+  id: string;
   kind: StageKind;
   name: string;
+  composite?: boolean; // true if stage uses data from other stages (e.g., reveal)
```
cjqian, line 38: Nit: `dependsOnOtherStages` or too long?

vivtsai: Hmm I don't have a strong preference, though I figured "composite" might be a faster term to refer to it (e.g., "composite stage") throughout the codebase (e.g., define it in documentation and then use everywhere)?

PR #139, opened 2024-07-18
Add payout stage to experiment

### LINKED ISSUE #25: Implement parameterized payout / incentives description and calculation
From the tech doc:
> Payment information for the user who did the experiment.
> Lucas: prototype something so we have a payment page. 
> Follow. General guidelines. ~10 EUR/hour.
> Have a %increase if you win (selected leader & for getting the right value).
> TODO([stefano.palminteri@gmail.com](mailto:stefano.palminteri@gmail.com) [Maël Lebreton](mailto:mael.lebreton@gmail.com)): formula for how much we pay participants.

Moving forward, we should think about how to calculate payouts as a general experiment feature (can we untangle this logic from the Lost at Sea-specific logic?)

vivtsai:
Adds backend definition and frontend UI views for initial version of PayoutStage. Fixes #25 

> NOTE: This also publishes all participant survey responses as public data (for use in payout and reveal stages).

Payout works as follows:
- Each payout stage contains 1+ payout groups (`PayoutBundle` items in code) that are added together for the final payout
- Within each payout group, experimenters can define 1+ payout items (`PayoutItem` items), which have fields for both fixed and per question amounts (e.g., one payout item might be $10 fixed amount + $2 per question = $14 if the participant gets 2 questions correct). This is also where experimenters can specify which survey stage questions to use for payout calculations and whether or not an election leader's (instead of current participant's) answers should be used.
- For each payout group, experimenters can either set the "strategy" to "add all payout items in the group" or "randomly select one payout item in the group"

Random selection strategies are defined via frontend config UI, but executed on the backend during experiment creation: a "scoring config" with selected payout items, etc. only is added to each PayoutStageConfig.

Note that this is just an initial implementation of payout; we may want to adjust based on evolving experiment designs / payouts in new experiments. Happy to add in-repo documentation on this once the implementation is stable.

Define payout groups of payout items

View payout stage


functions/src/endpoints/experiments.endpoints.ts
```diff
@@ -58,6 +65,52 @@ export const createExperiment = onCall(async (request) => {
 
       // Create the stages
       for (const stage of data.stages) {
+        // If payout stage, use payout config to generate scoring config
+        if (stage.kind === StageKind.Payout) {
+          const getScoringQuestion = (question: RatingQuestionConfig) => {
```
cjqian, line 70: This is great! Just a flag that we may want to generalize this moving forward (with different question types, such as multiple choice).

lit/src/shared/lost_at_sea/utils.ts
```diff
@@ -150,25 +194,30 @@ export function getRatingQuestionFromPair(
...
+  return stage.game === LAS_ID;
 }
 
 /**
- * Get Lost at Sea item ranking.
+ * Check if stage is a LostAtSea rating survey
  */
-export function getLostAtSeaItemRanking(item: ItemName) {
-  return ITEMS[item].ranking;
+export function isLostAtSeaSurvey(stage: StageConfig) {
```
cjqian, line 206: Maybe

return (isLostAtSeaGameStage) and (stage.kind is survey) ?

Otherwise, rename since this might not be the logic that's checked here

utils/src/types/items.types.ts
```diff
@@ -118,4 +118,21 @@ export const ITEMS: Record<ItemName, Item> = {
...
     ranking: 7,
   },
-};
\ No newline at end of file
+};
+
+/**
+ * Get Lost at Sea item item pair ranking answer.
+ */
+export function getLostAtSeaPairAnswer(item1: ItemName, item2: ItemName) {
```
cjqian, line 126: Nice, we can use this logic for multiple choice too (1 item has ranking 1, others 0).

Perhaps we can generalize the naming since this doesn't just apply to LostAtSea

cjqian (approved):
All nits, please feel free to merge! This looks great, I really like the way you designed the payout bundle since there are lots of hidden complexities.

utils/src/types/items.types.ts
vivtsai: Hmm right now the array used is specific to LAS, but I like the idea of generalizing it later. Maybe we can add a note to refactor it at that point?

cjqian: Sounds good, yes!

PR #141, opened 2024-07-19
Add fields to Experiment document

vivtsai:
New fields include:
- publicName (string, shown to participants)
- description (string)
- tags (list of strings)
- author (string, experimenter ID)
- starred (map from experimenter ID string to boolean isStarred)

This PR currently only updates the backend; frontend UI controls will be added to enable setting experiment description, tags, etc.
This also formats experiment cards to show date.


lit/src/services/experiment_service.ts
```diff
@@ -13,7 +13,7 @@ import { FirebaseService } from "./firebase_service";
 import { RouterService } from "./router_service";
 
 import { Snapshot } from "../shared/types";
-import {
+  import {
```
cjqian, line 16: Nit

cjqian (approved):

lit/src/components/header/header.ts
```diff
@@ -117,7 +117,7 @@ export class Header extends MobxLitElement {
       || activePage === Pages.PARTICIPANT_SETTINGS) {
       return "Settings";
     } else if (activePage === Pages.EXPERIMENT) {
-      return "My Experiment";
+      return this.experimentService.experiment?.name ?? "Experiment";
```
cjqian, line 120: Excellent

PR #143, opened 2024-07-22
Update UX for configuring experiment metadata

vivtsai:
- Enable experimenters to set public name, description for experiments
- Refactor metadata config UI
- Only show "delete experiment" buttons when experimenters click to "edit"
- Fix small UX bugs (e.g., flex-shrink, chip height)
- Add small visual cues to frontend (e.g., indicator for lobby stage, experiment selection in sidenav)


cjqian (approved):
LGTM!

PR #145, opened 2024-07-24
Implement dynamic lobbies MVP.

### LINKED ISSUE #133: Implement dynamic lobbies / multi-part experiments

cjqian:
Fixes #133

- [x] Participants can join an experiment by navigating to the experimenter page (not in experimenter auth)
- [x] Experimenters can transfer participants from the lobby to other experiments

### DIFF
--- a/functions/src/endpoints/experiments.endpoints.ts
+++ b/functions/src/endpoints/experiments.endpoints.ts
@@ -235,14 +235,14 @@ export const createParticipant = onCall(async (request) => {
 
     // Create a new participant document
     const participantRef = experimentRef.collection('participants').doc();
-    const participantData: ParticipantProfile = {
+    const participantData: ParticipantProfile = data.participantData || {
       publicId: participantPublicId(experimentData.numberOfParticipants),
-      currentStageId,
+      currentStageId: currentStageId,
       pronouns: null,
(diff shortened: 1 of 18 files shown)


cjqian:
Uploading Screen Recording 2024-07-24 at 6.48.44 PM.mov…

vivtsai (commented):
Hi, this is awesome! A few things:

- Small bug with creating participants when participantData is null --> see [fix](https://github.com/PAIR-code/llm-mediation-experiments/commit/905b42df9e834eef686092cc3ac86e0116e2b73f)
- Could we move the "join experiment" logic under ParticipantService instead of in the profile_preview UI component?
- Similarly, I'm wondering if some of the participant creation should be under ParticipantService on frontend and/or participant.endpoints on backend instead of ExperimenterService and experiment.endpoints, respectively - though I do see that they're more configuration-related than "participating in experiment." Maybe we leave as-is but add some comments to explain why it's located where it is?
- When creating the multi-group experiments, could we start the other experiments as empty so that participants joining from lobby are the only ones?
- I also feel like it might be a good time to add a isLobby boolean to experiments—just since we are starting to have more logic around it (e.g., if isLobby, allow transfer of participants)

PR #146, opened 2024-07-25
Add discussed changes to lobby transfer implementation

vivtsai:
Changes discussed in #145:

https://github.com/user-attachments/assets/43e329ea-ee39-4e0a-b698-7f6d6b7d85fa

- create/delete functions have been refactored to services instead of UI components
- added isLobby field to experiment
- added transferConfig to experiment to specify the transfer experiment/participant IDs
- added button for participant to self-diretc themselves to the new experiment (once the experimenter has assigned it)
- deleteParticipant function is still present in experimenterService, but not called
- transferred participants are sorted at the bottom of experiment_preview page

Note that the following change has not yet been implemented:
- when creating multi-group (lobby) experiments, start all experiments with 0 participants (as lobby participants will self-join and non-lobby experiment participants will be assigned by experimenters)

### DIFF
--- a/functions/src/endpoints/experiments.endpoints.ts
+++ b/functions/src/endpoints/experiments.endpoints.ts
@@ -241,7 +241,7 @@ export const createParticipant = onCall(async (request) => {
     const participantRef = experimentRef.collection('participants').doc();
     const participantData: ParticipantProfile = {
       publicId: participantPublicId(experimentData.numberOfParticipants),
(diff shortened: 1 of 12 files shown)


cjqian (approved):
Wooohoooo amazing changes 💯💯💯💯

I will add the conditional start experiment logic with nThreshold and other changes discussed today.

PR #149, opened 2024-07-25
Make small group experiment UX changes

cjqian:
* Fix rendering progress bar in groups
* Add an option to wait for all participants to join when nMaxParticipants specified (checkbox in experiment config)
* Add relevant properties to transfer dashboard

Feel free to merge if it looks good!

### DIFF
--- a/functions/src/endpoints/experiments.endpoints.ts
+++ b/functions/src/endpoints/experiments.endpoints.ts
@@ -45,7 +45,7 @@ export const createExperiment = onCall(async (request) => {
     const document = app.firestore().collection(data.type).doc();
 
     await app.firestore().runTransaction(async (transaction) => {
-      const { name, publicName, description, tags, isLobby, group, numberOfParticipants, numberOfMaxParticipants } = data.metadata;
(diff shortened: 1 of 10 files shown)


vivtsai (commented):
This is fine with me for now, but it might be worth having an option to preview or view data from the pre-transfer (lobby) experiment (e.g., experimenters might want to look back on why they assigned specific participants to specific experiments). With this change there's now no easy way to route to the pre-transfer info—but we can also update that later :)

vivtsai:
> This is fine with me for now, but it might be worth having an option to preview or view data from the pre-transfer (lobby) experiment (e.g., experimenters might want to look back on why they assigned specific participants to specific experiments). With this change there's now no easy way to route to the pre-transfer info—but we can also update that later :)

Another option might be to keep the old preview (of lobby experiment), with a second button to route to the new experiment page (or directly into the preview as you have it now).

vivtsai (approved):

cjqian:
Good idea, added a second icon in the transfer case :)

PR #165, opened 2024-07-29
Update latest instructions and limit transfers to max experiments.

### LINKED ISSUE #160: Don't allow transfer to full experiments

cjqian:
Fixes #160 #132


vivtsai (commented):
Fixed bug with adding final stages—which was actually a bug regarding participant transfer (we want pre-lobby stages, answers, participant ID, etc. to be copied over to the post-lobby experiment).

PR #193, opened 2024-08-01
Adds configurable attention check module and fix some post-playtest bugs.

### LINKED ISSUE #177: Add custom Prolific codes for different cases

cjqian:
Fixes #177 #190 #187

### DIFF
--- a/lit/src/app.ts
+++ b/lit/src/app.ts
@@ -94,7 +94,7 @@ export class App extends MobxLitElement {
       case Pages.PARTICIPANT_STAGE:
         if (
           !this.authService.isExperimenter &&
-          this.experimentService.experiment?.attentionCheckParams
+          this.experimentService.experiment?.attentionCheckParams?.waitSeconds
         ) {
           return html` <attention-check-popup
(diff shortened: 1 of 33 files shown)


lit/src/app.ts
```diff
@@ -92,11 +92,15 @@ export class App extends MobxLitElement {
       case Pages.PARTICIPANT:
         return this.renderParticipantView(this.renderParticipant());
       case Pages.PARTICIPANT_STAGE:
-        // TODO: Ignore this for experimenters, and allow experimenters to add this.
-        if (false) {
+        if (
+          !this.authService.isExperimenter &&
```
vivtsai, line 96: What do you think about verifying "isExperimenter" (and maybe the attention check params?) inside the attention-check-popup component (so that it's not possible to render it for experimenters viewing things)? Just to specifically safeguard the popup against experimenters in preview mode (since we could use it in other locations later).

vivtsai: We can follow up with this refactor though!

cjqian: Done!

Issue #215, opened 2024-08-05
Add a note / or a feature to contact the experimenters on prolific


cjqian:

vivtsai:
Help text tooltip can be used for this.

PR #216, opened 2024-08-05
Fix WTL implementation bug, TOS rendering bug, and adjust LAS Prolific settings

### LINKED ISSUE #214: Fix drag-and-drop issue with election (unable to select)

vivtsai:
[x] Filtering ranking answers on the frontend prevents the full ranking list from being written on the backend, which in turn impacts which participants can be ranked (e.g., currenly unranked participants) on the frontend.
[x] Implement WTL algorithm
[x] Fixes the TOS rendering issue (double HTML conversion)
[x] Minor cosmetic changes, adding a "Contact experimenters" note (and attention check 90 -> 120)

Fixes #214 #213 #215 #216


functions/src/triggers/stages.triggers.ts
```diff
@@ -149,17 +149,47 @@ export const publishStageData = onDocumentWritten(
...
+        const stagesSnapshot = await app
+          .firestore()
+          .collection(`experiments/${experimentId}/stages`)
+          .get();
+        const wtlStageDoc = stagesSnapshot.docs
+          .map((doc) => doc.data())
+          .find((data) => data.kind === StageKind.WTLSurvey);
+
+        if (wtlStageDoc) {
+          const wtlDoc = await app
```
vivtsai, line 167: Can you get the wtlDoc from the stagesSnapshot.docs above? Rather than calling Firestore again?

functions/src/triggers/stages.triggers.ts
```diff
@@ -149,17 +149,47 @@ export const publishStageData = onDocumentWritten(
...
+            .firestore()
+            .doc(`experiments/${experimentId}/publicStageData/${wtlStageDoc.id}`)
+            .get();
+          const wtlData = wtlDoc.data() as WTLSurveyStagePublicData;
+
+          // Get the participant IDs of the top two WTL scores.
+          const pScores = wtlData.participantScores;
+          const wtlCandidates = Object.entries(pScores)
+            .sort(([, scoreA], [, scoreB]) => scoreB - scoreA)
+            .slice(0, 2)
```
vivtsai, line 177: What happens if multiple participants have the same high score (e.g., everyone puts 10)? If it's fine to randomly pick, then can we add a comment clarifying this is WAI?

functions/src/triggers/stages.triggers.ts
```diff
@@ -149,17 +149,47 @@ export const publishStageData = onDocumentWritten(
...
         participantRankings[participantPublicId] = data.rankings;
 
-        // TODO: When determining current leader,
-        // only use participants whose scores are in the top two
-        // for willingness to lead (WTL).
-        //
-        // Determine top two scores from WTL, then filter participantRankings
-        // by participants who have those scores.
-        const currentLeader = getCondorcetElectionWinner(participantRankings);
+        let currentLeader = getCondorcetElectionWinner(participantRankings);
```
vivtsai, line 155: Instead of calling this early, could we only call getConcorcetElectionWinner once determining which participants should be sent in? E.g., updating participantRankings if needed and then call at the end

functions/src/triggers/stages.triggers.ts
```diff
@@ -149,17 +149,47 @@ export const publishStageData = onDocumentWritten(
...
+          const wtlCandidates = Object.entries(pScores)
+            .sort(([, scoreA], [, scoreB]) => scoreB - scoreA)
+            .slice(0, 2)
+            .map(([id]) => id);
+
+          // Update participant rankings
+          const initialRankings = publicData.participantRankings;
+          initialRankings[participantPublicId] = data.rankings;
+
+          const filteredRankings: Record<string, string[]> = Object.fromEntries(
```
vivtsai, line 184: Is there a reason you're using a ranking done by the participant? Could we instead just grab the list of participants from experiment metadata? Asking because it's possible a participant ranking e.g., publicData.participantRankings[publicId] could be incomplete if an attention check is failed, etc.

functions/src/triggers/stages.triggers.ts
cjqian: Done

functions/src/triggers/stages.triggers.ts
vivtsai: Wait, ignore. I read `initialRankings[participantPublicId] = data.rankings;` backwards (or wrong) oops

functions/src/triggers/stages.triggers.ts
cjqian: Done!

functions/src/triggers/stages.triggers.ts
cjqian: I tried and it didn't quite work; we can look into this during the refactor.

PR #228, opened 2024-09-24
Add UX bug fixes and features to refactored platform

cjqian:
Add bug fixes and features to improve UX following the refactored platform's first playtest.

### DIFF
--- a/frontend/src/components/header/header.ts
+++ b/frontend/src/components/header/header.ts
@@ -1,7 +1,7 @@
 import '../../pair-components/button';
 import '../../pair-components/icon_button';
 import '../../pair-components/tooltip';
-
+import '../../components/experiment_builder/experiment_builder_nav';
 import {MobxLitElement} from '@adobe/lit-mobx';
 import {CSSResultGroup, html, nothing} from 'lit';
 import {customElement} from 'lit/decorators.js';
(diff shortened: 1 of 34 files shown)


frontend/src/components/header/header.ts
```diff
@@ -137,12 +137,31 @@ export class Header extends MobxLitElement {
...
+
+        <pr-button
+          color="primary"
+          variant="tonal"
+          ?disabled=${!this.experimentEditor.canEditStages}
+          @click=${() => { this.experimentEditor.toggleStageBuilderDialog(true) }}
+        >
+          Load game
+        </pr-button>
+      </pr-tooltip>
```
vivtsai, line 164: Remove accidental `</pr-tooltip>`?

frontend/src/shared/cohort.utils.ts
```diff
@@ -26,19 +26,10 @@ export function getCohortDescription(cohort: CohortConfig) {
 export function getCohortParticipants(
   participants: ParticipantProfile[],
   cohortId: string,
-  countObsoleteParticipants = true, // participants who left the experiment
   includeParticipantsPendingTransferToCohort = true,
 ) {
   let participantList = participants;
 
-  // If not counting obsolete, filter out participants
```
vivtsai, line 34: What do you think about leaving this in? Even though it won't be applicable (since the field is not accessible via frontend), the field still exists in the backend data structure and someone might need it in the future?

utils/src/experiment.ts
```diff
@@ -31,8 +31,6 @@ export interface CohortParticipantConfig {
   minParticipantsPerCohort: number|null;
   // Max number of participants for experiment cohort (or null if no limit)
   maxParticipantsPerCohort: number|null;
-  // If false, exclude booted participant from min/max participant counts
-  includeAllParticipantsInCohortCount: boolean;
```
vivtsai, line 35: Do you think we might ever need this in the future? If so, it may be best to keep the field in (and just not use it / have it default to false) to avoid backwards compatibility issues later.

frontend/src/components/stages/stage_footer.ts
```diff
@@ -41,16 +42,11 @@ export class Footer extends MobxLitElement {
...
       await this.participantService.completeLastStage();
-      // Navigate to landing
-      this.routerService.navigate(
-        Pages.PARTICIPANT,
-        {
-          'experiment': this.routerService.activeRoute.params['experiment'],
-          'participant': this.routerService.activeRoute.params['participant'],
-        }
-      );
+      await this.participantService.routeToEndExperiment(ParticipantStatus.SUCCESS);
```
vivtsai, line 48: Just to confirm: we want to route participants to home (which will show them a 403 error since they don't have access) rather than the landing page for their current experiment (where they can still view the answers they submitted)?

frontend/src/components/header/header.ts
cjqian: Done, thank you!

utils/src/experiment.ts
cjqian: Sounds good, will leave this field in and just hide it in the experiment creation.

frontend/src/shared/cohort.utils.ts
cjqian: Done.

vivtsai (approved):

Issue #182, opened 2024-07-31
Add time-elapsed feature for each stage (start/end)


cjqian:
This will be helpful for better lobby timeouts.

vivtsai:
We have timestamps for stage completed. I can add some for stage started.

cjqian:
It's a little bit of finagling, but we can infer stage started from the last stage completed, so this is very P2.

I'm going to take this and fix the lobby elapsed logic.

PR #250, opened 2024-09-25
Fix lobby countdown logic, add boot participant confirmation, and improve experiment creation redirect.

### LINKED ISSUE #248: Add verification step before "booting"

cjqian:
Fixes #248 #182

### DIFF
--- a/frontend/src/components/stages/transfer_view.ts
+++ b/frontend/src/components/stages/transfer_view.ts
@@ -6,11 +6,16 @@ import {customElement, property, state} from 'lit/decorators.js';
 
 import {core} from '../../core/core';
 import {AuthService} from '../../services/auth.service';
+import {ExperimentService} from '../../services/experiment.service';
 import {ParticipantService} from '../../services/participant.service';
 import {
+  getTimeElapsed,
   ParticipantStatus,
-  TransferStageConfig
+  TransferStageConfig,
+  getRgbColorInterpolation,
 } from '@deliberation-lab/utils';
+
+import {getCurrentStageStartTime} from '../../shared/participant.utils';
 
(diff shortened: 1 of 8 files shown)


frontend/src/components/stages/transfer_view.ts
```diff
@@ -48,36 +54,57 @@ export class TransferView extends MobxLitElement {
 
     return html`
       <stage-description .stage=${this.stage}></stage-description>
-      ${this.renderCountdown()}
-      ${this.renderOverlay()}
+
+      <div>
+        Please wait while we transfer you to the next stage of the experiment.
```
vivtsai, line 59: Does it make sense to not hardcode text here since experimenters can customize it via the stage description?

frontend/src/components/stages/transfer_view.ts
```diff
@@ -48,36 +54,57 @@ export class TransferView extends MobxLitElement {
...
+    const seconds = this.timeRemainingSeconds % 60;
+
+    const timeColor = getRgbColorInterpolation(
+      '#A8DAB5', // Green
+      '#FBA9D6', // Red
+      this.timeRemainingSeconds,
+      this.stage?.timeoutSeconds
+    );
+    return html` <div>
+      If the transfer cannot be completed, the experiment will end in:
```
vivtsai, line 92: We should only render this message if `this.stage.enableTimeout` is true. (Actually, we should probably not render a countdown at all if timeout is not enabled.)

cjqian: Makes sense, changed. (This was the case before, but we were checking on `countdownSeconds` which is set if `enableTimeout`).

vivtsai (approved):

PR #253, opened 2024-09-26
Generalize election module to support items instead of participants, and add the initial gift card exchange game.

cjqian:


frontend/src/components/stages/election_editor.ts
```diff
@@ -0,0 +1,130 @@
...
+    `;
+  }
+
+  private renderElectionItems() {
+    if (!this.stage) return nothing;
+    const electionItems: ElectionItem[] = this.stage.electionItems || [];
+
+    const addItem = () => {
+      if (!this.stage) return;
+      const newId = String(electionItems.length + 1); // Generate a numeric ID as string
```
vivtsai, line 74: Can we use `generateId` from `utils/shared` or some other random method? This is going to generate duplicates if an item is deleted

vivtsai: could also add a `createElectionItem` function in `utils/stages/election_stage.ts`

frontend/src/components/stages/election_view.ts
```diff
@@ -94,18 +121,25 @@ export class ElectionView extends MobxLitElement {
...
   }
 
-  private renderDraggableParticipant(profile: ParticipantProfile) {
-    if (profile.publicId === this.participantService.profile?.publicId) {
-      return nothing;
-    }
+  private renderElectionItem(item: ElectionItem) {
+    // TODO: Allow item photos?
+    return html`
+        <div class="right">
```
vivtsai, line 127: Missing <div class="participant"> between line 126, 127

cjqian: Done, thank you!

PR #278, opened 2024-09-29
Update payout config and add payout editor

vivtsai:
Fixes #139 

payout editor


utils/src/stages/payout_stage.ts
```diff
@@ -26,11 +116,190 @@ export function createPayoutStage(
...
+/** Create default payout item. */
+export function createDefaultPayoutItem(
+  config: Partial<DefaultPayoutItem> = {},
+): DefaultPayoutItem {
+  return {
+    id: config.id ?? generateId(),
+    type: PayoutItemType.DEFAULT,
+    name: config.name ?? '',
+    description: config.description ?? '',
+    isActive: config.isActive ?? true,
```
cjqian, line 139: In what scenario would a payout item be created but not active?

cjqian (approved):

utils/src/stages/payout_stage.ts
vivtsai: Right now, for LAS either Part 2 XOR Part 3 is not active. This was the least disruptive way I could think of to allow LAS experimenters to quickly "randomize" from the payout editor (the LAS game config also randomly selects the isActive on creation). We can take it out if you think it's okay to randomize only on the backend (e.g., only add one of the two payout items)?

PR #280, opened 2024-09-30
Prevent non-creators from editing experiments

### LINKED ISSUE #266: Don't allow non-creators to delete experiments or edit experiment/stage configs
To start, disable the delete button in frontend.

To complete, check "if creator" in Firebase cloud functions so it can't go through even if somehow sent.

Finally, update wiki documentation AND note in experiment builder to explain permissions: Only creators can edit/delete the config. If you mark "public," anyone will be able to create cohorts/transfer participants (otherwise just you + any "readers" you add*)

*will file another bug for adding readers support

vivtsai:
Fixes #266 by allowing non-creators to preview, but not edit, experiment metadata settings and stage configs.

This also fixes bugs where some stage config fields were not disabled properly (e.g., when editing is disabled).


cjqian (approved):
We can merge this for now, but let's prioritize fixing the editing bug found this morning.

PR #283, opened 2024-10-02
Fix all unclaimed bugs from the previous deployment playtest.

cjqian:
Minus one (re: waiting on reveal stage dependencies), which we can file as a bug.

### DIFF
--- a/frontend/src/components/stages/chat_interface.ts
+++ b/frontend/src/components/stages/chat_interface.ts
@@ -41,7 +41,7 @@ export class ChatInterface extends MobxLitElement {
   private readonly participantService = core.getService(ParticipantService);
   private readonly routerService = core.getService(RouterService);
 
-  @property() stage: ChatStageConfig|undefined = undefined;
+  @property() stage: ChatStageConfig | undefined = undefined;
   @property() value = '';
   @property() disableInput = false;
   @property() showInfo = false;
@@ -49,9 +49,7 @@ export class ChatInterface extends MobxLitElement {
 
   private sendUserInput() {
     if (this.value.trim() === '') return;
-    this.participantService.createChatMessage(
-      { message: this.value.trim() }
-    );
+    this.participantService.createChatMessage({message: this.value.trim()});
     this.value = '';
   }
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/stages/ranking_editor.scss
```diff
@@ -49,4 +49,10 @@ pr-icon-button {
 .indented {
   /* width of checkbox plus gap */
   margin-left: calc(48px + common.$spacing-medium);
+}
+
+.warning{
+  font-style: italic;
+  color: var(--md-sys-color-error);
+  width: 80%;
```
vivtsai, line 57: Nit: use the common info-width variable (the same one used for stage descriptions, info/tos stages). I can push a fix for this.

vivtsai (approved):

vivtsai:
Thanks! I also left a comment (in doc) re: the reveal stage dependencies.

frontend/src/components/stages/survey_reveal_view.ts
```diff
@@ -178,14 +178,28 @@ export class SurveyReveal extends MobxLitElement {
...
+        (question as MultipleChoiceSurveyQuestion).options
+          .map((item) => item.text)
+          .join(', ');
+    }
+
     return html`
       <div class="table-row">
         <div class="table-cell number-row">${rowIndex}</div>
-        <div class="table-cell">${question.questionTitle}</div>
-        ${showAllParticipants
```
vivtsai, line 185: This was accidentally taken out. Putting it back in in #286

frontend/src/components/stages/chat_interface.ts
```diff
@@ -222,18 +223,29 @@ export class ChatInterface extends MobxLitElement {
...
-          || this.participantService.isReadyToEndChatDiscussion(this.stage.id, currentDiscussionId)}
+        ?disabled=${isDisabled}
         ?loading=${this.readyToEndDiscussionLoading}
         @click=${onClick}
       >
         Ready to end discussion
       </pr-button>
     `;
+    return isDisabled
+      ? html`<pr-tooltip text="You can move on once others are also ready to move on."
```
vivtsai, line 244: FYI: You can set pr-tooltip text="" if you want the tooltip to not show. So a cleaner version might be to always have the tooltip wrapped around the button, but conditionally render the tooltip text.

Issue #243, opened 2024-09-25
Clear experiment state when not on experiment page


cjqian:
Right now, experiment state persists when navigating out of one experiment and into another (unless the experimenter refreshes).

cjqian:
@vivtsai Please provide context or close it out if you don't remember what this is about either 😆

PR #286, opened 2024-10-02
Update firestore/storage loading and participant answer saving

vivtsai:
- Move calls to load (subscribe to) Firestore data (based on current experiment ID, participant ID, etc.) from components to RouterService. This fixes bugs with out-of-date data subscriptions #243 
- Create ParticipantAnswerService for storing participant stage answers on the frontend, then writing to Firestore on save or "next stage" only. This fixes latency/syncing bugs #265 
- Fix display bugs and add display enhancements
  - Render participant view at full height if not logged in as experimenter (i.e., no header banner visible)
  - Add chip for "public" experiments in home gallery
  - Fix undefined bug with rendering waiting stage message in participant preview header banner
  - Make nav headers sticky
- Fix image loading issues
  - Update Firebase storage rules to allow anyone to read
  - Create ImageService to handle loading images from Firebase storage (NOTE: in the future, load images per experiment)
- Fix frontend errors in prod (e.g., undefined objects)
- Refactor ranking stage logic (for best practices)
- Update new cohort's publicStageData on participant transfer


functions/src/participant.endpoints.ts
```diff
@@ -110,16 +111,49 @@ export const updateParticipant = onCall(async (request) => {
...
+        .doc(data.participantConfig.currentCohortId)
+        .collection('publicStageData')
+        .doc(stage.id);
+
+      switch (stage.kind) {
+        case StageKind.SURVEY:
+          const publicStageData = (await publicDocument.get()).data() as StagePublicData;
+          publicStageData.participantAnswerMap[publicId] = stage.answerMap;
+          transaction.set(publicDocument, publicStageData);
+        default:
```
cjqian, line 153: Is there any other potential public data, like chat data or ranking data?

cjqian (approved):

functions/src/participant.endpoints.ts
vivtsai: Right now, I don't think so: it wouldn't make sense for participants' previous chat messages or rankings to follow them into the new cohort (with different chat history / list of participants to rank).

Oh, although maybe if it's a non-election ranking, it would make sense to carry over rankings? Let me know what you think / happy to push an update here.

PR #288, opened 2024-10-02
Fix playtest issues with stage access, setup

### LINKED ISSUE #241: Reveal stage should only add options for surveys before it

cjqian:
Fixes #241 

* Disable editing "Wait for all participants" in reveal stage under specific conditions.
* Only allow revealing preceding stages (#241).
* Get rid of lock overlay on transfer stage.
* Font improvements.

### DIFF
--- a/frontend/src/components/progress/progress_stage_waiting.scss
+++ b/frontend/src/components/progress/progress_stage_waiting.scss
@@ -48,11 +48,11 @@ h2 {
 }
 
 .participant {
+  @include typescale.body-small;
   @include common.flex-column-align-center;
   flex-wrap: wrap;
   gap: common.$spacing-medium;
   max-width: 100px;
   overflow-wrap: anywhere;
   text-align: center;
-  font-size: var(--md-sys-typescale-body-small-font-size);
 }
--- a/frontend/src/components/progress/progress_stage_waiting.scss
+++ b/frontend/src/components/progress/progress_stage_waiting.scss
@@ -1,5 +1,5 @@
-@use "../../sass/common";
-@use "../../sass/typescale";
+@use '../../sass/common';
+@use '../../sass/typescale';
 
 :host {
   @include common.flex-column;
@@ -51,7 +51,8 @@ h2 {
   @include common.flex-column-align-center;
   flex-wrap: wrap;
   gap: common.$spacing-medium;
-  max-width: 80px;
+  max-width: 100px;
   overflow-wrap: anywhere;
   text-align: center;
+  font-size: var(--md-sys-typescale-body-small-font-size);
 }
--- a/frontend/src/components/stages/base_stage_editor.ts
+++ b/frontend/src/components/stages/base_stage_editor.ts
@@ -10,6 +10,9 @@ import {core} from '../../core/core';
 import {ExperimentEditor} from '../../services/experiment.editor';
 
 import {
+  ElectionStrategy,
+  RevealAudience,
+  RankingType,
   StageConfig,
   StageKind,
   StageProgressConfig,
@@ -24,7 +27,7 @@ export class BaseStageEditorComponent extends MobxLitElement {
 
   private readonly experimentEditor = core.getService(ExperimentEditor);
 
-  @property() stage: StageConfig|undefined = undefined;
+  @property() stage: StageConfig | undefined = undefined;
 
   override render() {
     if (this.stage === undefined) {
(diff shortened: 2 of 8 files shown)


frontend/src/components/stages/reveal_editor.ts
```diff
@@ -44,25 +42,58 @@ export class RevealEditor extends MobxLitElement {
...
+          (expStage) => expStage.id === stage.id
+        );
+        return stageIndex < currentStageIndex;
+      });
+
+      return filteredStages;
+    };
+
+    const stageOptions = getStageOptions();
+    if (stageOptions.length === 0) {
```
vivtsai, line 78: Instead of rendering twice, can we wrap the tooltip around the main return statement and set tooltip text to empty (which means no tooltip will render) if disabled?

frontend/src/components/stages/reveal_editor.ts
```diff
@@ -44,25 +42,58 @@ export class RevealEditor extends MobxLitElement {
...
     `;
   }
 
   private renderAddMenu() {
     if (!this.stage) return nothing;
 
-    const stageOptions = getStagesWithReveal(this.experimentEditor.stages).filter(
-      stage => stage.id !== this.stage?.id
-    );
+    // Get all stages with a reveal stage.
```
vivtsai, line 54: Nit: duplicate lines 54, 55

frontend/src/components/stages/reveal_editor.ts
```diff
@@ -44,25 +42,58 @@ export class RevealEditor extends MobxLitElement {
...
+    // Get all stages with a reveal stage.
+    const getStageOptions = () => {
```
vivtsai, line 56: Nit: Make this a utils function (as it's not component or state reliant)?

frontend/src/components/progress/progress_stage_waiting.scss
```diff
@@ -51,7 +51,8 @@ h2 {
   @include common.flex-column-align-center;
   flex-wrap: wrap;
   gap: common.$spacing-medium;
-  max-width: 80px;
+  max-width: 100px;
   overflow-wrap: anywhere;
   text-align: center;
-}
\ No newline at end of file
+  font-size: var(--md-sys-typescale-body-small-font-size);
```
vivtsai, line 57: Instead of setting font size, include a font bundle (which will also adjust line height, spacing, etc.), e.g., `@include typescale.body-small);`

frontend/src/components/stages/base_stage_editor.ts
```diff
@@ -129,16 +129,90 @@ export class BaseStageEditorComponent extends MobxLitElement {
     `;
   }
 
+  private shouldDisableWaitForAllParticipants(): boolean {
```
vivtsai, line 132: Can you make this a utils function that takes in list of stages, reveal stage ID?

frontend/src/components/stages/base_stage_editor.ts
```diff
@@ -129,16 +129,90 @@ export class BaseStageEditorComponent extends MobxLitElement {
...
-      this.experimentEditor.updateStage({ ...this.stage, progress });
+      const progress: StageProgressConfig = {
+        ...this.stage.progress,
+        waitForAllParticipants: !waitForAllParticipants,
+      };
+      this.experimentEditor.updateStage({...this.stage, progress});
     };
 
+    if (this.shouldDisableWaitForAllParticipants()) {
+      if (!waitForAllParticipants) {
```
vivtsai, line 189: I don't think it makes sense to update the wait bit on render (for instance, what if you never click on the stage?). Can we instead add the setting waitForAllParticipants logic when addStage (appending a new stage) or setStages (setting a list of stages) in experimentEditor is called?

frontend/src/components/stages/base_stage_editor.ts
```diff
@@ -129,16 +129,90 @@ export class BaseStageEditorComponent extends MobxLitElement {
...
-      const progress: StageProgressConfig = { ...this.stage.progress, waitForAllParticipants: !waitForAllParticipants };
-      this.experimentEditor.updateStage({ ...this.stage, progress });
+      const progress: StageProgressConfig = {
+        ...this.stage.progress,
+        waitForAllParticipants: !waitForAllParticipants,
+      };
+      this.experimentEditor.updateStage({...this.stage, progress});
     };
 
+    if (this.shouldDisableWaitForAllParticipants()) {
```
vivtsai, line 188: Instead of rendering twice (in this if statement and below), can we just have the main statement and conditionally set render the warning, set disabled, etc.?

vivtsai:
Thanks for the quick turnaround on these! I left some comments, mainly re: moving calculations that are not component/state dependent to utils (maybe `experiment.utils` or a new `stage.utils`), setting waitForAllParticipants logic in the correct spot, and combining "if" renders with default renders so that things like `<md-checkbox></md-checkbox` don't need to be set up twice/kept in sync.

frontend/src/components/stages/reveal_editor.ts
cjqian: Done.

frontend/src/services/experiment.editor.ts
```diff
@@ -112,30 +116,35 @@ export class ExperimentEditor extends Service {
...
-    const index = this.stages.findIndex(stage => stage.id === newStage.id);
+    const index = this.stages.findIndex((stage) => stage.id === newStage.id);
     if (index >= 0) {
       this.stages[index] = newStage;
     }
   }
 
   setStages(stages: StageConfig[]) {
-    this.stages = stages;
+    for (let stage of stages) {
```
vivtsai, line 134: set this.stages = [] first! I can push a fix for this

vivtsai: Also, my bad - I realized we actually need to do the mustWait whenever stages are edited as well. Pushing a fix for this.

vivtsai (approved):

PR #289, opened 2024-10-03
Add some clarity about mediators in chat.

cjqian:
Could be good to check that mediators work after this branch.


vivtsai:
Thanks! I'm closing this because I cherry-picked both commits into a new `mediator-viv` branch that includes more mediator enhancements.

Issue #285, opened 2024-10-02
Implement multi-agent conversation logic


cjqian:
Instead of looping through agents in `createMediatorMessage` trigger (meaning the first agent that has something to say will always have priority), add logic to allow the best agent response to be sent.

vivtsai:
@cjqian does random sampling fix this? Or are you adding more logic to choose "best response"?

PR #290, opened 2024-10-03
Add ability to configure JSON and show debugging for mediators

### LINKED ISSUE #284: Allow JSON response output in LLM prompts

vivtsai:
- Add a "debug" toggle button in experimenter header banner - when it's toggled on, the chat stage shows agents in the panel + explanations underneath LLM mediator messages; when it's off, the chat stage is identical to what participants see
- Add option to turn on "JSON mode" and specify which JSON fields should be used to populate the main chat message (defaults to "response") and explanation (defaults to "explanation")
- Add ability to edit prompt for existing mediators from experimenter right panel (also, add API key configuration editor to same panel)

Debugging mode with LLM mediator JSON responses/explanations

Chat editor with JSON configuration options

Edit prompt while in chat stage

Fixes #284

### DIFF
--- a/frontend/src/components/experimenter/experimenter_data_editor.ts
+++ b/frontend/src/components/experimenter/experimenter_data_editor.ts
@@ -32,6 +32,7 @@ export class ExperimenterDataEditor extends MobxLitElement {
       this.authService.writeExperimenterData({...data, apiKeys});
     };
 
+    // TODO: Make this more clear when this has been saved.
     return html`
       <div class="section">
         <div class="title">API keys</div>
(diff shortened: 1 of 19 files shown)


cjqian:
This fixes #[#285](https://github.com/PAIR-code/deliberate-lab/issues/285). Note that the random multi-agent sampling hasn't been tested. An easy way to test is to create mediators A, B, C who always respond A, B or C. In the current version, A will always respond. With this update, we'll hear from A, B, C randomly.

cjqian (approved):

PR #295, opened 2024-10-03
Add "anonymous profile" option for participant profiles

### LINKED ISSUE #294: Implement anonymous participants setting

cjqian:
This fixes #294. Now, experimenters can specify "Set anonymous profiles" in the profile editing stage. 

When participants are created, they are assigned an identity:
image

And will be shown this screen rather than the profile editing screen:
image

### DIFF
--- a/frontend/src/components/participant_profile/profile_avatar.ts
+++ b/frontend/src/components/participant_profile/profile_avatar.ts
@@ -7,6 +7,13 @@ import {customElement, property} from 'lit/decorators.js';
 import {classMap} from 'lit/directives/class-map.js';
 
 import {styles} from './profile_avatar.scss';
+import {getHashIntegerFromString} from '@deliberation-lab/utils';
+import {
+  MAN_EMOJIS,
+  WOMAN_EMOJIS,
+  PERSON_EMOJIS
+} from '../../shared/constants';
+import {getAvatarBackgroundColor} from '../../shared/participant.utils';
 
 /** Participant profile avatar */
 @customElement('profile-avatar')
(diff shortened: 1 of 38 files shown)


frontend/src/components/participant_profile/profile_avatar.ts
```diff
@@ -19,27 +21,35 @@ export class ProfileAvatar extends MobxLitElement {
   @property() disabled = false;
   @property() tooltip = '';
 
+  MAN_EMOJIS = ['👨🏻', '👨🏼', '👨🏽', '👨🏾', '👨🏿'];
```
vivtsai, line 24: move to `shared/constants.ts`

frontend/src/components/participant_profile/profile_viewer.ts
```diff
@@ -0,0 +1,54 @@
...
+          <div class="reveal">
+            <profile-avatar .emoji=${profile.avatar} .square=${true}>
+            </profile-avatar>
+            <div class="info">
+              <div class="title">${profile.name}</div>
+              <div class="subtitle">${profile.pronouns}</div>
+            </div>
+          </div>
+        </div>
+      </div>
```
vivtsai, line 45: Add stage_footer so participants can still continue

vivtsai (commented):
I can push these fixes!

vivtsai:
NOTE: Updated anonymous profile configuration so that the createParticipant endpoint only takes in one new parameter, `isAnonymous`. Based on this, we use functions in `utils/src/participant.ts` to populate the profile with animal name/avatar based on the participant's number (e.g., if there are animal options ["Lion", "Tiger", "Bear"], participant-0 gets "Lion", participant-2 gets "Tiger", participant-3 gets "Bear", participant 4 gets "Lion 1", etc.

vivtsai (approved):

Issue #114, opened 2024-07-01
Interface issue: how can people keep up with LLMs?


cjqian:
* LLM is really quick (perhaps solved with less LLM intervention)
* Outputs longer text than the humans (information ratio is so low, takes time to read it, lose track of the conversation)
* Maybe color-code threads by participant?

vivtsai:
Closing for now as we added typing delays.

Issue #101, opened 2024-07-01
Mediator appears to double-post


cjqian:
![Image](https://github.com/PAIR-code/llm-mediation-experiments/assets/4844633/d50bec10-790c-4331-ad31-d3ed0a77aa46)

vivtsai:
NOTE: Let's double-check, but I have a hunch this is because the LLM response to both participants A and B are the same (e.g., participant A posts, LLM response A loading, participant B posts, LLM response B loading... but since posting messages takes a while, both responses A and B post).

PR #310, opened 2024-10-18
Improve multi-agent chat, fix Prolific bugs, improve experiment creation.

### LINKED ISSUE #285: Implement multi-agent conversation logic

cjqian:
Fixes #285 #237 #299 #101 #114

* Agent changes: Fix multi-agent conversations, add agent mute feature, add experimenter instruction configuration field (and easier json/string understanding), simulated latency, and a reality TV discussion game to showcase changes.
* Fix Prolific integration and locked stages bugs.
* Add stage response previews to profile preview.
* Allow checkboxes to be unchecked.
* Clean up experiment creation: allow copying stages.


frontend/src/services/cohort.service.ts
```diff
@@ -111,7 +111,7 @@ export class CohortService extends Service {
   getUnlockedStageParticipants(stageId: string) {
     return this.getAllParticipants().filter(
       participant => isUnlockedStage(participant, stageId)
-        && !isObsoleteParticipant(participant)
+        || !isObsoleteParticipant(participant)
```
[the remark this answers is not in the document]
vivtsai: Update: I see the update! 
[cfbbc77 - Fix waiting stage hanging bug on obsolete participants.](https://github.com/PAIR-code/deliberate-lab/pull/310/commits/cfbbc772b18f51db44d6bd3de647cd59a5987498)

utils/src/stages/chat_stage.ts
```diff
@@ -26,6 +26,7 @@ export interface ChatStageConfig extends BaseStageConfig {
   kind: StageKind.CHAT;
   discussions: ChatDiscussion[]; // ordered list of discussions
   mediators: MediatorConfig[];
+  muteMediators: boolean; // Whether mediators are muted.
```
vivtsai, line 29: (No action required, just keeping track) This sets `EXPERIMENT_VERSION_ID = 2`

frontend/src/components/stages/survey_editor_menu.scss
```diff
@@ -4,6 +4,7 @@
 
 :host {
   @include common.flex-column;
+  margin-top: -30px;
```
vivtsai, line 7: Nit: Instead of doing this, can we just wrap the title and survey editor in a horizontal flex? (Otherwise, this creates a weird vertical flex layout because the survey editor menu gets squished.)

In survey_editor:
```
<div class="section">
  <div class="header">
    <div class="title">Survey questions</div>
    <survey-editor-menu></survey-editor-menu>
  </div>
  ...
</div>
```

utils/src/stages/survey_stage.ts
```diff
@@ -154,6 +155,7 @@ export function createCheckSurveyQuestion(
     id: config.id ?? generateId(),
     kind: SurveyQuestionKind.CHECK,
     questionTitle: config.questionTitle ?? '',
+    isRequired: config.isRequired ?? false,
```
vivtsai, line 158: (No action item, just keeping track) This sets `EXPERIMENT_VERSION_ID = 3`

frontend/src/components/stages/survey_reveal_view.ts
```diff
@@ -114,6 +114,10 @@ export class SurveyReveal extends MobxLitElement {
   }
 
   private renderIcon(correctAnswer: string, selectedAnswer: string) {
+    if (correctAnswer === '') {
+      return;
```
vivtsai, line 118: Nit: `return nothing`?

frontend/src/components/participant_profile/profile_preview.ts
```diff
@@ -40,6 +47,16 @@ export class Preview extends MobxLitElement {
       return nothing;
     }
 
+    this.participantService.updateForRoute(
```
vivtsai, line 53: Can you clarify what this is doing (and why it's called twice)?

frontend/src/components/popup/accept_transfer_popup.ts
```diff
@@ -56,6 +56,18 @@ class TransferPopup extends MobxLitElement {
   private handleYes() {
     this.analyticsService.trackButtonClick(ButtonClick.TRANSFER_ACCEPT);
     this.participantService.acceptParticipantTransfer();
+
+    // Move to next stage.
+    const nextStageId = this.experimentService.getNextStageId(
```
vivtsai, line 61: Can we have a param for moving to next stage? This works well if participants are sitting in a transfer stage, but not if, say, a participant is in a chat stage (but everyone else drops out) and the experimenter transfers them to a different cohort's chat.

vivtsai: Actually, a param might be complicated. My understanding is that we previously handled this by having it move to next stage if the participant is currently in a transfer stage, else don't move. Is there something wrong with that logic (i.e., maybe we could just fix/add that)?

frontend/src/components/header/header.ts
```diff
@@ -151,20 +151,35 @@ export class Header extends MobxLitElement {
     const activePage = this.routerService.activePage;
     const profile = this.participantService.profile;
 
-    const renderParticipantProfileBanner = () => {
+    const renderParticipantProfileBanner = (profile: any) => {
+      if (!profile) return;
+      
+      let detailText = '';
+      if (profile.currentStatus === ParticipantStatus.BOOTED_OUT) {
```
vivtsai, line 158: Nit: Use utils function to get detail text based on participant status?

vivtsai:
Sorry for the delay on this! Let's increment `EXPERIMENT_VERSION_ID = 3` (see resolved comments) based on the utils changes that were made (maybe we can also leave a comment there noting at which commit it was last incremented).

frontend/src/components/stages/survey_editor_menu.scss
cjqian: Wow, yes, done!

frontend/src/components/participant_profile/profile_preview.ts
cjqian: Yes! It's not called twice; we update the participantService and the participantAnswerService. We want to fetch the current survey answer data, which relies on participantAnswerService (which in turn relies on participantService), so the routes need to be updated. If this is not included, participantAnswerService will not be able to provide the current survey answer data.

frontend/src/components/popup/accept_transfer_popup.ts
cjqian: I don't think we have logic that moves participants to the next stage if they're in a transfer stage currently; I've implemented this here (only move to next stage if on a transfer stage).

cjqian:
All feedback has been addressed!

frontend/src/components/participant_profile/profile_preview.ts
vivtsai: Ah sorry, I missed the `answer` in `participantAnswerService`!

The answer service is used for saving participants' local edits, so I'm not sure it makes sense to be using it in the profile preview (as experimenters or anyone else on this page would only be viewing what's saved in Firebase). Can we load information from participantService (which should have an answer map) only?

vivtsai: Update: sorry, I just realized you're using the answer service because the stage view components rely on it. Let me think about this some more.

vivtsai: What do you think about updating these when the experimenter clicks on the current participant (rather than in the lit component)? That way, the participant answers are up-to-date on the frontend (based on the experiment manager's currently viewed participant) regardless of whether or not this component is shown.

cjqian: Sounds fine to me (esp. given the true purpose of the answer service); would you mind making this change? I'm not sure I follow entirely on the implementation.

vivtsai: Yes, I'll add it to this PR before merging!

vivtsai: FYI, this is the fix I pushed: https://github.com/PAIR-code/deliberate-lab/pull/310/commits/de96f3aeb50871a1e108f89ea5a2f5b6cb90389a

vivtsai (approved):

frontend/src/components/popup/accept_transfer_popup.ts
```diff
@@ -57,10 +57,14 @@ class TransferPopup extends MobxLitElement {
     this.analyticsService.trackButtonClick(ButtonClick.TRANSFER_ACCEPT);
     this.participantService.acceptParticipantTransfer();
 
-    // Move to next stage.
-    const nextStageId = this.experimentService.getNextStageId(
-      this.participantService.profile?.currentStageId!
-    )!;
+    // Move to next stage if on a transfer stage.
```
vivtsai, line 60: @cjqian what do you think about moving this routing logic to acceptParticipantTransfer (called a few lines up where we already have "set current stage to next if transfer stage" logic? Happy to make the change in a new branch/PR, or if you're working on anything similar, feel free to take it

vivtsai: Update: I can take this. It might be related to the transfer stage refresh bug I'm currently investigating.

PR #337, opened 2024-11-15
Make attention checks and fails experimenter-triggered

vivtsai:


cjqian:
Note: We want to remove the attention check configuration from the experiment metadata as well!

cjqian (approved):

cjqian:
Small notes:

1. The tooltip on the participant bar isn't showing for me on Chrome (works for the cohort bar)
2. Future work: It'd be nice to have a "time elapsed since attention check sent"

PR #350, opened 2024-12-10
Connect hosted LLM 

### LINKED ISSUE #305: Add a Llama API key option
* Expand the APIKeyConfig: https://github.com/PAIR-code/deliberate-lab/blob/0004ad9dea299522013538b15cec994852491107/utils/src/experimenter.ts#L21
* Modify the logic to check whether _any_ API key exists, rather than depending only on the Gemini key
* Change the "Create experimenter" view to include a Llama option:
* Add the implementation e.g. https://github.com/PAIR-code/deliberate-lab/blob/main/functions/src/api/gemini.api.ts

dimits-ts:
Fixes #305

Allow open-source LLMs hosted on any server to be called as Gemini would.

Changes:
* Abstracted LLM API type 
    * Created an interface to use any kind of API key 
    * Exported interface for LLM responses regardless of API
    * API validity checks can now be extended for any kind of API
* Added Ollama server support
     * Supports most open source models (is not limited to LLaMa)
     * Added frontend section to connect to Ollama server through HTTP URL
     * Added sample Ollama server in the form of a [docker container](https://github.com/dimits-ts/deliberate-lab-utils/tree/master/llm_server) for testing
* Included unit tests for most new functions   

TODOs:
* LLaMa server may need some kind of authentication
* While Ollama supports almost all popular open-source LLMs, this patch only allows the use of llama3.2. 
     * We can include a text area or dropdown menu in Settings with all available models. The issue is that doing so would necessitate pulling and deploying all such models on the remote Ollama server first.
* API key check only checks whether the key is not empty, not if it's valid. This also goes for Gemini keys. 
* Right now all moderators are either Gemini or LLaMa. Future versions may include mixing different kinds of models (see LLaMa integration design doc)


cjqian:
@mkbehr FYI!

functions/src/api/llama.api.ts
```diff
@@ -0,0 +1,116 @@
+/**
+ * Client handling communications with an Ollama server.
+ * 
+ * Code assumes that the LLaMa instance is hosted on the provided IP address,
+ * and is managed through the `ollama` framework (https://github.com/ollama/ollama).
+ * Example docker instance hosting an ollama server: https://github.com/dimits-ts/deliberate-lab-utils/tree/master/llm_server
+ * 
+ * Note: there already exists a client library for JavaScript, but not for Typescript.
```
mkbehr, line 8: I'm working on adding some more API support, and I'm planning to support Ollama either via the openai library and Ollama's openai API compatibility, or via multi-llm-ts. This looks good in the meantime, just a heads up, and I wouldn't recommend putting a lot more work into polishing this.

functions/src/api/llama.api.ts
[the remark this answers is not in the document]
dimits-ts: You are correct, this was a holdout from when I wasn't sure which backend library was going to be used for hosting the OSS LLMs. I will correct it.

frontend/api/gemini.api.ts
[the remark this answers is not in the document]
vivtsai: Agreed - I think only having API functions in one place would be best. @dimits-ts is there a reason to make LLM calls from the frontend? Our initial assumption when setting up the codebase was that backend calls would suffice/be preferred.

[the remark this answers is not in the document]
vivtsai: At quick glance it looks like the `frontend/api` is no longer used and could be deleted (since the actual ollama call uses `functions/src/api`)?

frontend/api/llama.api.test.ts
```diff
@@ -0,0 +1,19 @@
+import { OllamaChat } from "./llama.api";
```
cjqian, line 1: Amazing, thank you for adding tests!

frontend/api/llama.api.ts
```diff
@@ -0,0 +1,153 @@
...
+ * 
+ * Note: there already exists a client library for JavaScript, but not for Typescript.
+ */
+
+/**
+ * The JSON schema of Ollama LLM responses.
+ */
+type IncomingMessage = {
+    model: string,
+    created_at: Date,
```
cjqian, line 14: Nit: Consider renaming to dateCreated for consistency with the rest of the codebase

frontend/api/llama.api.ts
```diff
@@ -0,0 +1,153 @@
...
+    message: LlmMessage,
+    done_reason: string,
```
cjqian, line 16: Nit: For this variable and others, rename using mixedCase (e.g. doneReason, modelType) instead of done_reason, model_type for consistency.

frontend/src/components/stages/chat_editor.ts
```diff
@@ -39,17 +40,17 @@ export class ChatEditor extends MobxLitElement {
 
     // Check if experimenter's API key exists
     let apiCheck;
-    if (!this.authService.experimenterData?.apiKeys.geminiKey) {
+    if (!checkApiKeyExists(this.authService.experimenterData)) {
       apiCheck = html`
         <div class="warning">
-          <b>Note:</b> In order for LLM calls to work, you must add your Gemini
-          API key under Settings.
+          <b>Note:</b> In order for LLM calls to work, you must add your Gemini API key
```
cjqian, line 46: Let's make this more generic: "you must add an API key or server configuration under Experimenter Settings." (Here and in chat_panel.ts)

cjqian (changes requested):

frontend/api/gemini.api.ts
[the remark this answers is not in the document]
dimits-ts: You are both correct, I have no idea how the files ended up getting duplicated. The actual API calls are made to the functions module, the frontend/api module is unused. I am removing it, sorry for the confusion.

frontend/api/llama.api.ts
dimits-ts: This was an outdated version of the file which I mistakenly left committed (see mike's comment above). Sorry for the confusion!

frontend/src/components/stages/chat_editor.ts
dimits-ts: Fixed!

frontend/api/gemini.api.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (error.message.includes(QUOTA_ERROR_CODE.toString())) {
+      console.log("API quota exceeded");
+    } else {
+      console.log("API error");
+    }
+    console.log(error);
+  }
+
+  // Log the response
+  console.log(response);
```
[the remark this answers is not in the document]
dimits-ts: I removed most from the functions module and my own code in chat_triggers, but left the warnings and errors as console.warn and console.error respectively. I didn't touch the logs for the rest of the codebase, let me know if I should.

.vscode/settings.json
mkbehr, line 1: This probably shouldn't go in the repository.

functions/src/api/llama.api.ts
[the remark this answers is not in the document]
mkbehr: I'm still seeing some references to "llama", mostly the things defined in utils/src/experimenter.ts: llamaApiKey, LlamaServerConfig, LLAMA_CUSTOM_URL.

functions/src/api/llama.api.ts
mkbehr: The openai API will work fine wherever ollama is hosted. That method won't touch openai's servers or anything, it's just a client library that we don't need to maintain ourselves.

utils/src/experimenter.ts
```diff
@@ -14,15 +13,40 @@ export interface ExperimenterProfile {
 
 /** Experimenter data (written to Firestore under experimenterData/{id}). */
 export interface ExperimenterData {
+  /* 
+  Currently supports either all Llama or all gemini
+  TODO: refactor this as a list of types and values for each mediator(see design document)
+  */
   id: string;
-  apiKeys: APIKeyConfig;
```
vivtsai, line 18: @dimits-ts @mkbehr @cjqian The original idea here was to have APIKeyConfig contain all API key info (as ExperimenterData may eventually contain non-API key settings). What do you think about keeping APIKeyConfig and nesting all api settings under there? E.g.,

```
APIKeyConfig {
  geminiApiKey: string;
  llamaApiKey: LlamaServerConfig;
  activeApiKeyType: ApiKeyType;
}
```

(we also may want `geminiApiKey` to point to a config, but I'm fine with waiting until we have something else to store there)

vivtsai (changes requested):
Hi @dimits-ts - thanks for working on this! I left a comment about structuring ExperimenterData / API key configs.

.vscode/settings.json
cjqian: +1

cjqian:
I took my final pass and pushed a small cosmetic change. I've also verified that Gemini keys continue to work following this change.

 Dimitris, this looks phenomenal! Thank you for all of your hard work on this and this incredible contribution.

Leaving to @vivtsai to merge when she's done with her final pass.

functions/src/api/llama.api.ts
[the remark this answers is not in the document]
dimits-ts: Ah right, I overlooked these. There should be no references to llama anywhere now. Apologies.

.vscode/settings.json
dimits-ts: Fixed and added to gitignore

dimits-ts:
I have refactored the types as @vivtsai recommended, and took the opportunity to make the code in experimenterData.ts more DRY-like. The changes seem to be compatible with the new changes in the upstream.

vivtsai (approved):

cjqian (approved):

vivtsai:
@dimits-ts thanks so much for working on this! quick heads up that after merging, I updated the ApiKeyType enum strings to be all-caps for consistency: #385

PR #373, opened 2024-12-18
Fix chip negotiation payout/offer calculations and hardcoded chip values

### LINKED ISSUE #365: Update payout to amount over baseline bonus

cjqian:
Fix #365

### DIFF
--- a/frontend/src/components/stages/payout_summary_view.ts
+++ b/frontend/src/components/stages/payout_summary_view.ts
@@ -30,11 +30,6 @@ import {
   calculatePayoutResult,
   calculatePayoutTotal,
 } from '@deliberation-lab/utils';
-import {
-  N_INITIAL_BLUE_CHIPS,
-  N_INITIAL_GREEN_CHIPS,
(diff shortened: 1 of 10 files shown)


frontend/src/components/stages/payout_summary_view.ts
```diff
@@ -112,15 +116,28 @@ export class PayoutView extends MobxLitElement {
...
 
   private renderChipPayoutItemResult(
     item: ChipPayoutItemResult,
     currency: PayoutCurrency
   ) {
+    const getInitialTotal = () => {
+      let total = 0;
+      for (const result of item.chipResults) {
+        let quantity = 0;
+        if (result.chip.name.includes('red')) {
```
vivtsai, line 130: can we add a TODO to generalize this later? the initial chips should really be based on the chip stage's configuration (i.e., ChipStageConfig has a ChipItem list that specifies the starting quantities), not custom variables

vivtsai (approved):

PR #381, opened 2024-12-20
Add basic OpenAI API text completion support.

mkbehr:
Allow Deliberate Lab to talk to any model endpoint that supports the OpenAI text completion API.

This is configured by setting environment variables on the functions emulator:
- OPENAI_BASE_URL: Controls whether to use the OpenAI API, and the base URL to use. If empty, use the Gemini API instead.
- OPENAI_API_KEY: API key. May be empty, e.g. for local Ollama.
- OPENAI_MODEL_NAME: Model name. May be empty, e.g. for local servers that only serve one model.

For example, to use an API endpoint that requires an API key but not a model name:

`OPENAI_BASE_URL=http://hostname:port/v1 OPENAI_API_KEY=key firebase emulators:start --import ./emulator_test_config`

Once #350 is merged, we can use its configuration interface to configure this API.


cjqian:
Looks good and minimally invasive as-is; happy to merge pending @vivtsai's review.

Side note: The console logging for APIs is getting a little out of hand (not your fault, this is the way we do things currently)! I've filed https://github.com/PAIR-code/deliberate-lab/issues/382 as an idea on how to surface this to the experimenter.

functions/src/api/openai.api.ts
```diff
@@ -0,0 +1,79 @@
+import OpenAI from "openai"
+
+const MAX_TOKENS_FINISH_REASON = "length";
+
+export async function callOpenAITextCompletion(
+  apiKey: string,
+  prompt: string,
+  modelName: string
+) {
+  console.log('callOpenAITextCompletion')
```
cjqian, line 10: Remove?

mkbehr: You meant the console.log line, right? Done. (Also changed the one below to console.error)

vivtsai (approved):

PR #386, opened 2025-01-09
Fix negotiation game playtest bugs

cjqian:


frontend/src/services/experiment.manager.ts
```diff
@@ -504,6 +504,8 @@ export class ExperimentManager extends Service {
     };
 
     this.updateParticipant(config);
+
+    // TODO: Handle if they're on the chip stage.
```
vivtsai, line 508: This should ideally be handled in the backend (should be solved by this month's endpoint refactor).

vivtsai (approved):

PR #392, opened 2025-01-10
Add configuration for OpenAI-compatible API.

mkbehr:
Move config options for the OpenAI-compatible API into the web interface, instead of requiring environment variables.

The base URL and API key can be configured from the experimenter interface, along with other API keys.

The chat stage includes configuration for model name, sampling parameters, and custom request body fields. Currently these only do anything for the OpenAI-compatible API, but we'll want to extend at least model name and sampling parameters to Gemini as well. The new config is displayed regardless of API choice, with a note that it only applies to the OAI API.


frontend/src/components/stages/chat_editor.ts
```diff
@@ -200,6 +200,34 @@ export class ChatEditor extends MobxLitElement {
...
+          model,
+        },
+        index
+      );
+    };
+
+    return html`
+      <div class="question-label">Model</div>
+      <div class="description">
+        Model ID for the agent. TODO: Currently only used for OpenAI and compatible APIs.
```
vivtsai, line 218: Note: The pr-textarea component has a "label" property where the description can be passed in

frontend/src/components/stages/chat_editor.ts
```diff
@@ -347,6 +378,193 @@ export class ChatEditor extends MobxLitElement {
...
+       );
+     };
+
+    const updateField = (fieldIndex: number, field: Partial<{ name: string; value: string }>) => {
+      this.updateAgent(
+        {
+          ...agent,
+          generationConfig: {
+            ...agent.generationConfig,
+            customRequestBodyFields: agent.generationConfig.customRequestBodyFields.map((f, i) =>
```
vivtsai, line 531: Instead of reassigning every single field, maybe just update the one at the given fieldIndex?

functions/src/stages/chat.triggers.ts
```diff
@@ -310,17 +311,14 @@ async function hasEndedChat(
   return false;
 }
 
-async function getAgentResponse(data: ExperimenterData, prompt: string): Promise<ModelResponse> {
+async function getAgentResponse(data: ExperimenterData, prompt: string, agent: AgentConfig): Promise<ModelResponse> {
```
vivtsai, line 314: Heads up: This file has been refactored to a separate agents.utils file. (Sorry!)

functions/src/stages/chat.triggers.ts
```diff
@@ -335,6 +333,18 @@ async function getGeminiResponse(data: ExperimenterData, prompt: string): Promis
   return await getGeminiAPIResponse(data.apiKeys.geminiApiKey, prompt);
 }
 
+async function getOpenAIAPIResponse(
```
vivtsai, line 336: I would also move this function to the new agent.utils file (basically any function that's not specifically tied to chat stage)

utils/src/experimenter.ts
```diff
@@ -34,15 +34,22 @@ export interface ExperimenterData {
 
 export interface APIKeyConfig {
   geminiApiKey: string, // distinct types since we don't want to lose information when switching between them
+  openAIApiKey: OpenAIServerConfig
```
vivtsai, line 37: WDYT about providing backwards compatibility on the frontend by setting `openAIApiKey` when `APIKeyConfig` is loaded by snapshot listeners (I believe in admin service and auth service)?

Another option could be to make `openAIApiKey` optional? (`OpenAIServerConfig|null` might also work but I'm not positive.)

Basically, just want to make sure we don't disrupt API key configs for people currently using the Google deployment.

utils/src/stages/chat_stage.ts
```diff
@@ -98,15 +98,31 @@ export interface AgentMediatorChatMessage extends BaseChatMessage {
   explanation: string;
 }
 
+
+export interface CustomRequestBodyField {
```
vivtsai, line 102: WDYT about moving these new interfaces (since they are related to agents, not specifically chat) to a new `utils/src/agent.ts` file?

(We will eventually move current agent logic in chat stage up to the experiment level, so it might make sense to put any new stuff in the "right" place.)

vivtsai: (We can also do this later - no worries if you want to just get this in first!)

utils/src/experimenter.ts
```diff
@@ -103,6 +111,13 @@ export function checkApiKeyExists(experimenterData: ExperimenterData | null | un
     return experimenterData.apiKeys.geminiApiKey !== INVALID_API_KEY
   }
 
+  if (experimenterData.apiKeys.activeApiKeyType === ApiKeyType.OPENAI_API_KEY) {
+    // A custom server could require no API key, and the default OpenAI server
+    // requires no base URL setting, but leaving both blank is invalid.
+    return (experimenterData.apiKeys.openAIApiKey.apiKey !== INVALID_API_KEY ||
```
vivtsai, line 117: This should be `&&`, not `||`, right?

frontend/src/components/stages/chat_editor.ts
```diff
@@ -349,6 +380,205 @@ export class ChatEditor extends MobxLitElement {
...
+
+
+     return html`
+       <div class="question-label">Custom request body fields</div>
+       <div class="description">
+         Add custom fields to the request body.
+       </div>
+       ${agent.generationConfig.customRequestBodyFields.map((field, fieldIndex) => html`
+         <div class="custom-field">
+           <pr-textarea
```
mkbehr, line 554: This setup ends up putting the name, value, and delete button all on different lines. I'd rather have one line per field but wasn't sure how to do that.

frontend/src/components/stages/chat_editor.ts
vivtsai: Oh, good point. Let's leave as-is for now :)

utils/src/experimenter.ts
vivtsai: Ah, excuse the brain fart on my end. I think the comment is fine as is, though what probably would've clicked for me is something like "Custom server is valid if either API key or base URL is valid" (i.e., using "or" to match the actual statement)

utils/src/stages/chat_stage.ts
vivtsai: Thanks! Can we also export the new agent.ts in `utils/src/index.ts`?

frontend/src/components/stages/chat_editor.ts
vivtsai: Ah, it looks like the `custom-field` class you are using here isn't actually defined in the [chat_editor stylesheet](https://github.com/PAIR-code/deliberate-lab/blob/main/frontend/src/components/stages/chat_editor.scss).

You could define it in chat_editor.scss as:

```
.custom-field {
    @include common.flex-row; // arranges items in a horizontal line
    gap: common.$spacing-medium; // spacing between items
}
```
Note that that's in SASS as per some of our definitions in [common.scss](https://github.com/PAIR-code/deliberate-lab/blob/main/frontend/src/sass/_common.scss). In raw CSS, it would be:

```
.custom-field {
    display: flex; /* this defaults to flex-direction: row; you would additionally add flex-direction: column; for vertical arrangement */
    gap: 8px;
}
```

vivtsai: One thing we do for some similar items in survey_editor is add a bit of background/padding for each item, e.g.:

.custom-field {
    @include common.flex-row;
    background: var(--md-sys-color-surface-variant);
    border-radius: common.$spacing-small;
    gap: common.$spacing-medium;
    padding: common.$spacing-medium;
}

utils/src/experimenter.ts
vivtsai: Oh - we'll need a question mark after `openAIApiKey` now that it's optional, right?

functions/src/agent.utils.ts
```diff
@@ -32,6 +34,18 @@ export async function getGeminiResponse(data: ExperimenterData, prompt: string):
   return await getGeminiAPIResponse(data.apiKeys.geminiApiKey, prompt);
 }
 
+async function getOpenAIAPIResponse(
+  data: ExperimenterData, model: string, prompt: string, generationConfig: GenerationConfig
+): Promise<ModelResponse> {
+  return await getOpenAIAPITextCompletionResponse(
+    data.apiKeys.openAIApiKey.apiKey,
```
vivtsai, line 41: Add question marks after `openAIApiKey` (with fallback for `apiKey` if `openAIApiKey` doesn't exist)?

utils/src/experimenter.ts
vivtsai: Just kidding - I see the check you added above :)

vivtsai (changes requested):

vivtsai (approved):

utils/src/stages/chat_stage.ts
vivtsai: Committed change.

vivtsai:
Committed the requested changes to this branch and marked resolved.

NOTE: Left the chat_editor styling thread open in case @mkbehr wants to give it a go; if not, I'm happy to include the UI change in an upcoming "update experimenter dashboard UX" PR.

PR #428, opened 2025-01-22
Add endpoint to test agent participant prompts

vivtsai:
This allows experimenters to use a frontend button to call an agent participant testing endpoint:

image

The endpoint then constructs a prompt based on the current stage (TODO: add these prompts! right now it just builds a dummy prompt), sends to API, and console logs the response:

image

As we complete bugs like #413, we can fill in the TODO area by calling new utils/functions to construct the correct prompt.


cjqian (approved):
Cool!

PR #429, opened 2025-01-30
Update agent participant endpoint ranking example to fetch relevant F…

vivtsai:
Updated console output for ranking stage:

Screenshot 2025-01-29 at 6 26 43 PM

### DIFF
--- a/functions/src/stages/ranking.utils.ts
+++ b/functions/src/stages/ranking.utils.ts
@@ -1,0 +1,74 @@
+import {
+  ParticipantProfileExtended,
+  ParticipantStatus,
+  RankingStageConfig,
+  RankingStageParticipantAnswer,
+  createAgentParticipantRankingStagePrompt,
+  createRankingStageParticipantAnswer
+} from '@deliberation-lab/utils';
+import {getAgentResponse} from '../agent.utils';
+
+import * as admin from 'firebase-admin';
+import * as functions from 'firebase-functions';
+import { onCall } from 'firebase-functions/v2/https';
+
+import { app } from '../app';
+
+/** Use LLM call to generate agent participant response to ranking stage. */
+export async function getAgentParticipantRankingStageResponse(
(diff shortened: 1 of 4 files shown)


functions/src/stages/ranking.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    const participantAnswer = createRankingStageParticipantAnswer({
+      id: stage.id,
+      rankingList
+    });
+    console.log(
+      '✅ RankingStageParticipantAnswer\n',
+      JSON.stringify(participantAnswer)
+    );
+  } catch (error) {
+    console.log(error);
```
cjqian, line 70: Maybe a little more detail to wrap the error (e.g. "Fetching response from ranking")? Up to ou

functions/src/stages/ranking.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+export async function getAgentParticipantRankingStageResponse(
+  experimentId: string,
+  experimenterData: ExperimenterData, // for making LLM call
+  participant: ParticipantProfileExtended,
+  stage: RankingStageConfig
+) {
+  // Get list of public IDs for other participants who are active in the cohort
+  // (in case the agent participant is ranking participants)
+  // TODO: Use shared utils to determine isActiveParticipant
+  const activeStatuses = [
```
cjqian, line 27: Something to consider in case we want to keep it flexible - ask it to rank everybody, and then filter to active statuses? Up to you

cjqian (approved):

functions/src/stages/ranking.utils.ts
vivtsai: Let's skip for now as this is just an example (the ranking is not even happening now - it's just alphabetizing) to show how to set this up for other stages. There's a TODO to update the prompt; I can also add a TODO here to update logic accordingly.

PR #430, opened 2025-01-30
Enable random selection of payout items (per participant)

### LINKED ISSUE #422: Randomly select payout items per participant

vivtsai:
Fixes #422 #398 

When configuring payout items, experimenters can assign a "random selection ID" to each item. For each unique "random selection ID," one of the items tagged with that ID is randomly selected (for each participant as the participant profile is created).

> Example: Payout stage includes a payout item "survey-1" and a payout item "survey-2". Both payout items are assigned the random selection ID "survey-payout." When a participant joins the experiment, either survey-1 or survey-2 is selected (randomly), and a resulting map is written to the participant's stage data, e.g., `{ 'survey-payout': 'survey-1' }`. This map is then referenced when determining whether to calculate payout (e.g., if an item has the random selection ID 'survey-payout' but is not 'survey-1', it will not be included in payout).

The function `calculatePayoutResult` (used to render participant payout views and fill data download CSV columns) takes in an extra `PayoutStageParticipantAnswer` param and only constructs "payout item results" for items that are active AND (have no random selection ID XOR have a random selection ID and are selected as per the participant answer).

NOTE: This also fixes CSV comma/column bugs by running text values through a `toCSV` function.

### DIFF
--- a/frontend/src/components/participant_view/participant_view.ts
+++ b/frontend/src/components/participant_view/participant_view.ts
@@ -230,7 +230,8 @@ export class ParticipantView extends MobxLitElement {
         `;
       case StageKind.PAYOUT:
         return html`
-          <payout-participant-view .stage=${stage}></payout-participant-view>
+          <payout-participant-view .stage=${stage} .answer=${answer}>
+          </payout-participant-view>
         `;
       case StageKind.REVEAL:
         return html`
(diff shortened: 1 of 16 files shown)


cjqian (approved):
Cool! Is this the only thing that was programmatically defined in the shared/game stages files?

vivtsai:
> Cool! Is this the only thing that was programmatically defined in the shared/game stages files?

yes!

PR #442, opened 2025-02-04
Update stage waiting logic to rely on new stageUnlockMap field

### LINKED ISSUE #357: Fix waiting stage / participant progress / ready to end chat bugs
The number of participants may be inaccurate / get stuck. This resolves with refreshing. A current workaround is that we added "Hey please refresh your screen" to the waiting stage.

Another note / this showed up in the MIT playtests: Participants are able to access the negotiation stage while some folks (player-1) are on the waiting stage.

vivtsai:
Fixes #357

### DIFF
--- a/frontend/src/services/cohort.service.ts
+++ b/frontend/src/services/cohort.service.ts
@@ -58,6 +58,7 @@ export class CohortService extends Service {
 
   @observable experimentId: string|null = null;
   @observable cohortId: string|null = null;
+  @observable cohortConfig: CohortConfig|null = null;
 
   // Participants currently in the cohort
   @observable participantMap: Record<string, ParticipantProfile> = {};
@@ -246,6 +247,8 @@ export class CohortService extends Service {
     return this.chipLogMap[stageId] ?? [];
   }
 
+  // Called from participant service on participant snapshot listener
+  // (i.e., when participant's current cohort may have changed)
   async loadCohortData(experimentId: string, id: string) {
     if (id === this.cohortId) {
       return;
@@ -257,10 +260,33 @@ export class CohortService extends Service {
     this.unsubscribeAll();
 
     // Subscribe to data
+    this.loadCohortConfig();
     this.loadPublicStageData();
     this.loadChatMessages();
     this.loadChipLogEntries();
     this.loadParticipantProfiles();
+  }
+
+  /** Subscribe to current cohort config. */
+  private loadCohortConfig() {
+    if (!this.experimentId || !this.cohortId) return;
+
+    this.isCohortConfigLoading = true;
+    this.unsubscribe.push(
+      onSnapshot(
+        doc(
+          this.sp.firebaseService.firestore,
+          'experiments',
+          this.experimentId,
+          'cohorts',
+          this.cohortId,
+        ),
+        async (doc) => {
+          this.cohortConfig = doc.data() as CohortConfig;
+          this.isCohortConfigLoading = false;
+        }
+      )
+    )
   }
 
   /** Subscribe to public stage data. */
@@ -455,6 +481,7 @@ export class CohortService extends Service {
     this.unsubscribe = [];
 
     // Reset stage configs
+    this.cohortConfig = null;
     this.participantMap = {};
     this.chatMap = {};
     this.chatDiscussionMap = {};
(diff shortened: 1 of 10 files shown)


frontend/src/components/participant_view/participant_view.ts
```diff
@@ -179,16 +175,15 @@ export class ParticipantView extends MobxLitElement {
...
       !this.authService.isExperimenter
     ) {
       return html`
         <div class="content">
           <div>Stage not available yet</div>
           <pr-button @click=${() => {
-            this.participantService.setCurrentStageView(
-              this.participantService.profile?.currentStageId ?? undefined
-            );
+            this.participantService.setCurrentStageView(undefined);
```
cjqian, line 186: Would it make sense to not render the "Go to current stage" button at all in this case?

frontend/src/services/cohort.service.ts
```diff
@@ -261,12 +241,38 @@ export class CohortService extends Service {
...
+        doc(
+          this.sp.firebaseService.firestore,
+          'experiments',
+          this.experimentId,
+          'cohorts',
+          this.cohortId,
+        ),
+        async (doc) => {
+          this.cohortConfig = {
+            stageUnlockMap: {}, // for backwards compatibility
```
cjqian, line 267: Idea for future backend changes - in protos, they keep fields as existent but deprecated until the next major version change. Not sure how to incorporate this immediately but something to consider

frontend/src/shared/participant.utils.ts
```diff
@@ -139,13 +139,12 @@ export function isParticipantEndedExperiment(participant: ParticipantProfile) {
 }
 
 /** If participant is on or past the given stage. */
+// NOTE: "completedWaiting" map is now used to track when a participant reaches
```
cjqian, line 142: When I read `completedWaiting`, it's not clear to me that this implies that they finished the stage ID that is the key; rather, it seems that they have _completed the waiting stage_ for the stage ID that is the key. Is that what is meant?

utils/src/participant.ts
```diff
@@ -59,6 +59,9 @@ export interface ProgressTimestamps {
   // Stage ID to time that stage was marked completed
   completedStages: Record<string, UnifiedTimestamp>;
   // Stage ID to time that stage's waiting phase was marked completed
+  // NOTE: This field is now being used to track time that the participant
+  // arrived at the stage. Waiting is now "completed" when the stage (for
```
cjqian, line 63: At some point, perhaps we should refactor all of these into a timestamps wrapper (including startTimestamp, endTimestamp, etc.) so that it's easy to understand (e.g. timestamps.completedStages[''])

cjqian (approved):

frontend/src/components/participant_view/participant_view.ts
vivtsai: I realized there's an edge case: If you haven't clicked "start experiment," your current stage is inaccessible. Setting "undefined" takes you to the "landing view," which either shows the "start experiment" button, an "experiment is over" message, or routes you to the current page.

So basically this always routes you to the landing, which then routes you accordingly if needed.

I can leave a comment in the code about this.

frontend/src/services/cohort.service.ts
vivtsai: Yeah that would be cool! Although this is a little trickier because in this case we're adding a field.

utils/src/participant.ts
vivtsai: I'm not 100% sure what you mean by that? Right now, you would call participants.timestamps.completedStages['id'] which seems pretty similar?

frontend/src/shared/participant.utils.ts
vivtsai: Right now, `completedWaiting['my-stage']` means that you are ready to start `my-stage` (e.g., you either see the stage or a waiting screen).

We can rename the field to `readyToStart['my-stage']` or something like that. I just left comments for now because the field is otherwise unused and renaming it would mean that all past experiments wouldn't load properly (because missing the map).

Maybe I should take the chance to rename now and just prioritize an experiment version conversion option (e.g., via upload)? Let me know what you think.

vivtsai: Update: I think it's fine to rename now with `readyToStart` as an empty map on snapshot load for backwards compatibility. Let me do that (and we can continue to figure out the right prioritization for version conversion).

Issue #282, opened 2024-09-30
Add a linter


cjqian:

vivtsai:
And/or an editor config? https://editorconfig.org/

PR #451, opened 2025-02-12
Set up pre-commit Git hook for running Prettier and ESLint

vivtsai:
#282


mkbehr (approved):
Looks good! Can we also document somewhere what the style standards are and how to run the tools manually?

vivtsai:
> Looks good! Can we also document somewhere what the style standards are and how to run the tools manually?

Yes, good point! Let me put something in the wiki really quickly now and make sure it's documented in the new docs (that I'm about to start).

Issue #397, opened 2025-01-13
Improve agent conversation templating of what is sent to the API query


cjqian:
@mkbehr FYI

vivtsai:
Is this a duplicate of #396?

PR #494, opened 2025-04-04
Support structured output configuration.

mkbehr:
#397

- Experimenters can now configure custom schemas for structured outputs.
- Three special fields are supported: a field for the message, a field for whether or not to respond, and a field for an explanation or chain of thought. Experimenters can configure what these fields are named.
- The prompt will include a description of the output schema. Experimenters can disable this if they'd rather supply their own output examples.
- Experimenters can constrain the sampler to output valid json, or to output valid json in the specified schema. This is only supported for the Gemini API so far.
- Experiments with the old isJSON config should still work, but setting that config in new experiments is disabled.
- New experiments have a default config with structured outputs enabled, a premade schema with the three special fields, schema prompting, and no output constraints.
- All schemas must be flat. Nested objects and arrays are supported in the backend, but don't have UI support.
- The UI has some quirks: elements jump around when checkboxes are checked/unchecked, dropdown menus for field datatypes are too tall, and fields can't be reordered.
- I haven't written documentation yet.

- [x] Tests pass (except utils/src/experimenter.test.ts, which was already failing)

### DIFF
--- a/frontend/src/components/experiment_builder/agent_editor.ts
+++ b/frontend/src/components/experiment_builder/agent_editor.ts
@@ -15,13 +15,10 @@ import {ExperimentService} from '../../services/experiment.service';
 import {
   AgentChatPromptConfig,
   AgentPersonaConfig,
-  AgentResponseConfig,
   ApiKeyType,
   StageConfig,
   StageKind,
-  checkApiKeyExists,
-  DEFAULT_JSON_FORMATTING_INSTRUCTIONS,
-  DEFAULT_STRING_FORMATTING_INSTRUCTIONS,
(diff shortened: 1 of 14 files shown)


vivtsai:
@mkbehr thanks! I've made a new `structured-output` branch in this repository with these changes and am adding a few more:

- Resolving the broken test in experimenter
- Enabling backwards compatibility re: new StructuredOutputConfig
- Fixing the UI issues you mentioned
- Adding the description of this PR as documentation under the feature log)

Going to close this out for now and open a new PR for `structured-output` (with your changes + additions). Please feel free to review there!

PR #507, opened 2025-05-01
Add presence detection using the firebase realtime database websocket

### LINKED ISSUE #506: When a participant disconnects, I don't have any way to know
Our experiment involves actively matching participants for a chat conversation. Sometimes one of them will disconnect while awaiting transfer. When this happens, we have no way of knowing, and might end up transferring someone into a chat with a person who isn't there.

This happens, for example, when someone gets tired of waiting, closes their laptop, or returns the study on Prolific. But all we care about is whether the user actually has the window open. Attention checks help with this, but aren't great since we need to run them proactively.

We plan to automate this transfer process for a larger experiment, at which point this will become even more important.

(PR incoming)

raindrift:
Fixes #506

With this PR, whether a user has their browser window open and connected is reflected in the UI, and the `connected` property on their participant record. It allows the experimenter to know if the participant is still present, which is important for experiments that involve participants interacting with one another.

This works by using the firebase rtdb's existing presence functionality, and a cloud function that mirrors the user's state into firestore.

- [x] Tests pass

### DIFF
--- a/firestore/database.rules.json
+++ b/firestore/database.rules.json
@@ -1,0 +1,12 @@
+{
+  "rules": {
+    "status": {
+      "$experimentId": {
+        "$publicId": {
+          ".read": "true",
+          ".write": "true"
+        }
+      }
+    }
+  }
+}
--- a/utils/src/participant.ts
+++ b/utils/src/participant.ts
@@ -35,6 +35,7 @@ export interface ParticipantProfile extends ParticipantProfileBase {
   currentStatus: ParticipantStatus;
   timestamps: ProgressTimestamps;
   anonymousProfiles: Record<string, AnonymousProfileMetadata>;
+  connected: boolean;
 }
 
 /** Anonymous profile data generated from profile set. */
@@ -148,6 +149,7 @@ export function createParticipantProfileExtended(
     currentStatus: config.currentStatus ?? ParticipantStatus.IN_PROGRESS,
     timestamps: config.timestamps ?? createProgressTimestamps(),
     anonymousProfiles: {},
+    connected: false,
   };
 }
 
--- a/firebase.json
+++ b/firebase.json
@@ -22,9 +22,15 @@
   "storage": {
     "rules": "firestore/storage.rules"
   },
+  "database": {
+    "rules": "firestore/database.rules.json"
+  },
   "emulators": {
     "auth": {
       "port": 9099
+    },
+    "database": {
+      "port": 9000
     },
     "functions": {
       "port": 5001
--- a/frontend/src/app.ts
+++ b/frontend/src/app.ts
@@ -21,6 +21,7 @@ import {AuthService} from './services/auth.service';
 import {HomeService} from './services/home.service';
 import {Pages, RouterService} from './services/router.service';
 import {SettingsService} from './services/settings.service';
+import {PresenceService} from './services/presence.service';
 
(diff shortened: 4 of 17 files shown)


raindrift:
FYI, I have tested this on the emulator, where it works well and is quite fast. However, if we find in production that it's too slow to respond when someone connects, we should be able to add an endpoint so the client can update the status proactively, rather than waiting for the trigger to fire.

firestore/database.rules.json
```diff
@@ -0,0 +1,12 @@
+{
+  "rules": {
+    "status": {
+      "$experimentId": {
+        "$publicId": {
+          ".read": "true",
+          ".write": "true"
```
pdarche, line 7: Does this make the db open to the public?

pdarche (approved):
Looks good to me!

firestore/database.rules.json
raindrift: It makes this key open to the public, but you can only access records for known IDs. Since research participants are anonymous (from an auth perspective) there is no other way.

However, I notice that I misnamed that ID. So lemme go fix that...

utils/src/participant.ts
```diff
@@ -36,6 +36,7 @@ export interface ParticipantProfile extends ParticipantProfileBase {
   currentStatus: ParticipantStatus;
   timestamps: ProgressTimestamps;
   anonymousProfiles: Record<string, AnonymousProfileMetadata>;
+  connected: boolean;
```
vivtsai, line 39: Since this field is added, can we make the ParticipantProfile backwards compatible on the frontend (i.e., if reading in previously stored experiments) by adjusting the snapshot listeners in [ParticipantService](https://github.com/PAIR-code/deliberate-lab/blob/ae929e51d342df67d0e4c938413d63c37cd0ab08/frontend/src/services/participant.service.ts#L237), [ExperimentManager](https://github.com/PAIR-code/deliberate-lab/blob/ae929e51d342df67d0e4c938413d63c37cd0ab08/frontend/src/services/experiment.manager.ts#L421), and [CohortService](https://github.com/PAIR-code/deliberate-lab/blob/ae929e51d342df67d0e4c938413d63c37cd0ab08/frontend/src/services/cohort.service.ts#L485) (if field is meant to be included in ParticipantProfile, not just ParticipantProfileExtended - see other comment)? You'll notice we did this already with `agentConfig` field.

vivtsai, line 39: Is this new field intentionally under ParticipantProfile (all participants have access to other participants' connected status), or should it be under ParticipantProfileExtended (only experimenters or the current participant has access to the participant's status)? My understanding is that it might be more appropriate under the latter?

vivtsai (changes requested):

utils/src/participant.ts
raindrift: Oh good point! I wasn't considering backward compatibility at all.

raindrift: It wasn't intentionally there, but I think it needs to be there. With this change, `isActiveParticipant` depends on the `connected` property, and it's used in a bunch of the participant views.

raindrift:
Ok, updated to make the connected property optional. Also now makes sense for agent participants. Previously it would show them as always offline. I guess this is technically true in a way, but not useful. Now it will show them as always online instead.

PR #508, opened 2025-05-02
Fix OpenAI-compatible API and enable sampler constraints.

mkbehr:
The OpenAI-compatible API now supports constraining the sampler ("Structured output type") in the same way as the Gemini API.

Also fix a bug introduced in #500, where responses weren't being properly parsed from the chat completions API.

Also set default temperature to 1.0 from 0.7, to support recent OpenAI models.

- [x] Tests pass


functions/src/api/openai.api.test.ts
```diff
@@ -1,49 +1,158 @@
...
+  StructuredOutputDataType,
+  createStructuredOutputConfig,
+} from '@deliberation-lab/utils';
 import {getOpenAIAPIChatCompletionResponse} from './openai.api';
 import {ModelResponse} from './model.response';
 
+const DEFAULT_GENERATION_CONFIG: ModelGenerationConfig = {
+  maxTokens: 300,
+  stopSequences: [],
+  temperature: 0.7,
```
vivtsai, line 16: Do we want to change default temperature as discussed offline?

vivtsai (approved):

PR #509, opened 2025-05-08
Refactor experiment editor UI (layout, text fields)

vivtsai:


cjqian:
Some notes playing around with this:
* Remind me - does "Limit cohort to maximum number of participants" just "lock" the cohort at that number automatically? 

Experiment creation page:
* Visual consistency: The "Enable prolific" checkbox conditionally brings down the relevant input; can the "Limit cohort to maximum number of participants" above have the same behavior to keep the page as clean as possible on default?
* Bug: I was able to create an experiment with empty required fields (e.g. to reproduce - click "Enable prolific" and create. Might be worth checking on the validation logic?

Stage pages:
* What are your thoughts on keeping name and description outside of the accordion, and having "Popup metadata" and "Progression settings" as separate accordions underneath (default to closed)?

Agent creation:
* What happens if I save an experiment with a mediator without a chat stage? Could be confusing.

vivtsai:
> * Remind me - does "Limit cohort to maximum number of participants" just "lock" the cohort at that number automatically?
Yes, it does. We can add a tooltip to help clarify maybe?

> * Visual consistency: The "Enable prolific" checkbox conditionally brings down the relevant input; can the "Limit cohort to maximum number of participants" above have the same behavior to keep the page as clean as possible on default?
If we were to hide the "cohort max participants" field, then it automatically disappears when you hit Backspace inside the field (because that sets the max participants to 0/null, which is jarring). That's why I changed it to just disabled when the checkbox isn't checked.

> * Bug: I was able to create an experiment with empty required fields (e.g. to reproduce - click "Enable prolific" and create. Might be worth checking on the validation logic?
That's WAI, but we can adjust validation logic if you'd like. Right now we don't have much validation logic so the "required" things are mainly to signal the user's attention. I can also take off the "required" from those fields for now.
 
> * What are your thoughts on keeping name and description outside of the accordion, and having "Popup metadata" and "Progression settings" as separate accordions underneath (default to closed)?
Yes, that change is actually pending! Just wanted to start moving the fields first :)
 
> * What happens if I save an experiment with a mediator without a chat stage? Could be confusing.
This is also pending - I'm thinking we can put a warning in (and also have a quick link from the mediator editing page to add a chat stage).

cjqian:
Screenshot 2025-06-03 at 4 29 09 PM

One counter-intuitive behavior I noticed here - in the mediator panel, you can add _any_ stage. If you add a non-chat stage from this panel, it doesn't show up. Perhaps we can filter to "Mediator-enabled stages" or add more info / guidance here?

cjqian:
UI nit: Maybe add a "Stages" section header to make clear that the "Group discussion", "Group chat" are stages? And perhaps the "Add stage" could to to the right of that.
image

cjqian (approved):
Added some comments in the conversation. LGTM upon addressing!

vivtsai:
Thanks! Will adjust "add stage" and stage section header as you suggested.

> We allow you to select both "Wait for all active participants to reach this stage before allowing progression" and "Wait for a fixed number of participants to reach this stage before allowing progression" - is this working as intended?
Yes, those two things are slightly different. For instance, you could say "wait for all active participants" AND "wait for at least 3 participants." Then if 4 people are in the cohort, it will wait for 4. If 2 people are in the cohort, it will wait for 3.

vivtsai:
image

FYI, updating UX for clarifying "stage prompts" and how to add chat stages.

PR #530, opened 2025-05-30
Responsive participant view

### LINKED ISSUE #529: Make participant view responsive (mobile-friendly)
Our study is fairly large, and involves pulling from a participant pool (pro-life Prolific users) which isn't that much larger, so we need to be able to access as much of it as possible.

Although Prolific users tend to be on desktop at a higher rate than most people, lots of people still use phones/tablets. So it's important for the participant view to be available on mobile. Otherwise we risk not being able to recruit enough of the available participants to get the statistical power we need for our study.

(PR forthcoming)

raindrift:
Fixes #529

Adds some simple responsive layout to the participant view. On smaller screens:

- Hides the left sidebar showing the current stage
- Hides the left sidebar in the chat stage
- Shows chat participants and time remaining at the top of the window
- Puts the chat info inside the chat, so it scrolls off with new messages
- Reformats the buttons in the transfer popup

This code is a bit messy since I'm trying to get it out the door asap for an experiment we are running. But it does make participation possible on a phone. Especially, there is some duplicated code between `chat_panel` and `chat_interface` that could be refactored into a set of common components.

- [X] Tests pass
- [X] Appropriate changes to documentation are included in the PR

### DIFF
--- a/frontend/src/components/participant_view/participant_nav.scss
+++ b/frontend/src/components/participant_view/participant_nav.scss
@@ -1,6 +1,6 @@
-@use "../../sass/colors";
-@use "../../sass/common";
-@use "../../sass/typescale";
+@use '../../sass/colors';
+@use '../../sass/common';
+@use '../../sass/typescale';
 
 :host {
   @include common.flex-column;
@@ -47,7 +47,7 @@
   border-radius: common.$spacing-medium;
 
   &.experimenter-only {
-    opacity: .25;
+    opacity: 0.25;
   }
 }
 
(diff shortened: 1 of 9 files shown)


raindrift:
Note: This PR has some code that depends on types added in #528. So that one should be merged first. Or if it's removed, the code for disabling the next button around line 455 of `chat_interface.ts` should be removed. I would have chained this PR but I'm not quite sure how to do that when the branch I'd be chaining against lives in my fork.

raindrift:
Totally, that seems like a good way forward. It's pretty rough since I was
trying to get an experiment launched under deadline. If I were in your
position I wouldn't merge it straight into main. :)

raindrift:
Ok, I did a little refactor to move the duplicated timer code into the chat stage utils.

PR #532, opened 2025-06-03
Add cohort editing panel and persona-based agent participant configuration to experiment dashboard

vivtsai:
Fixes #<issue_number_goes_here>

> It's a good idea to open an issue first for discussion.

- [ ] Tests pass

### DIFF
--- a/firestore/firestore.rules
+++ b/firestore/firestore.rules
@@ -94,6 +94,11 @@ service cloud.firestore {
         allow write: if false; // Use cloud functions
       }
 
+      match /logs/{logId} {
+        allow read: if isExperimenter();
+        allow write: if false; // Use cloud functions
+      }
+
       match /stages/{stageId} {
         allow read: if true; // Public read
         // Experimenters can use `writeExperiment`, `deleteExperiment`
@@ -145,7 +150,7 @@ service cloud.firestore {
             allow write: if false; // Write using cloud functions
           }
 
-          match /logs/{logId} {
+          match /logs/{logId} { // TODO: Refactor chip logs
             allow read: if true;
             allow write: if false; // Write using cloud functions
           }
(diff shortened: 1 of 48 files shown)


cjqian:
1. Should we rename "Agent mediators" to "Agent chat mediators" or something to make the difference more clear? I'd imagine that an experimenter might not know the difference.
2. Thoughts on this mentioned in the previous PR (adding a "Stages" header, only being able to add chat stages)
3. Change wording (Specify agent mediator, delete agent mediator) for the Agent participant stage
4. Consider adding instructions e.g. "These are stage-specific instructions that you can customize. You will be able to customize an agent persona upon instantiation in an experiment" 
[image
](https://screenshot.googleplex.com/BazXbnSHEktipi9.jpg)

cjqian:
A few notes from playing around with this:

[] Currently, you can only pause responses from agent mediators (not participants). When I added multiple mediators, only one had the pause button. All agents should be able to be paused (and perhaps a global "end conversation" button for experimenters to disable the chat for all?)
[] Forking an experiment should fork mediators and participants as well
[] Surface the prompts for each participant more visibly (currently, it's in the JSON dump)
[] In my case, I was trying to set up a negotiation with participants P1, P2, and P3. I got tired of copying this into the agent persona prompt, so I tried to add it in the conversation instructions "Your name is P1". In retrospect, this makes sense that the agent participant didn't understand this in the profile stage. However, we should then be able to set specific instructions for each stage (not just the group chat), right?
[] Consider adding warnings if agent participants are created without custom instructions for that stage.
[] Consider adding a warning if an API key is not set yet while agent mediators / agent participants are being edited.. or maybe like a red notification on the key symbol in the left tab if there are no active API keys?

cjqian (approved):

PR #535, opened 2025-06-10
Enable MVP automatic cohort transfer based on survey answers

vivtsai:
This merges basic functionality as implemented by @raindrift from Plurality Institute in PR #524 with some small adjustments:
- Fix bug in test game's chat prompt config (merge conflict)
- Bundle all fields for automatic cohort transfer (survey ID, question ID, answer ID to participant count mappings) into an AutoTransferConfig (to make additional transfer types, e.g., by number only or by profile stage, easier to add later)
- Add placeholder text to frontend editing fields

Note that we'd eventually like to complete the following (in subsequent PRs):
- Update the frontend editor to enable users to select the survey stage and question ID and fill in number fields for each answer's participant count (similar to how payout for survey stage can currently be configured)
- Add "default auto transfer" (specify number of participants)
- Change frontend editor to use Material 3 textarea instead of pr-textarea (to match new experiment editor styles)
- Add indicator in experiment dashboard that transfer will happen automatically
- Add alerts (to alert panel and upcoming logging panel) when transfer fails
- Enable agent participants to wait for and accept transfers
- Check that the specified survey stage happens BEFORE the transfer stage
- Carry over the default "cohort participant config" from the main experiment config when initializing auto transfer settings


vivtsai:
Thanks! Will merge this into a temporary branch (not main) and open another PR to address comments.

PR #543, opened 2025-06-20
Refactor Firestore trigger functions to reduce latency

vivtsai:
Organize trigger-related logic into several helper functions and then only call a trigger for each document path once (rather than having multiple triggers for the same path).


functions/src/agent_participant.utils.ts
```diff
@@ -0,0 +1,186 @@
...
+        experimenterData,
+        participant,
+        stage,
+      );
+      answerDoc.set(rankingAnswer);
+      await completeStage();
+      participantDoc.set(participant);
+      break;
+    case StageKind.SURVEY:
+      if (!experimenterData) {
```
cjqian, line 131: Is this console logging statement the kind of thing that will eventually be put in a transaction log?

functions/src/agent_participant.utils.ts
```diff
@@ -0,0 +1,186 @@
...
+    // If participant is NOT agent, do nothing
+    if (!participant?.agentConfig) {
+      return;
+    }
+
+    // Otherwise, accept terms of service and start experiment
+    if (!participant.timestamps.startExperiment) {
+      participant.timestamps.startExperiment = Timestamp.now();
+    }
+    if (!participant.timestamps.acceptedTOS) {
```
cjqian, line 167: Is TOS in all experiments?

cjqian (approved):
Nothing actionable, just a few small comments. Thanks for doing this!

functions/src/agent_participant.utils.ts
vivtsai: The stage may not be in all experiments, but every participant has a "terms of service" bit, so for agent participants we just always mark it as "accepted" at the start.

PR #555, opened 2025-06-30
Reorganize platform homepage and experiment dashboard

vivtsai:


cjqian:
What other start chips do we envision (other than create an experiment)?

vivtsai:
> What other start chips do we envision (other than create an experiment)?

Quick group chat with mediator and transfer stage, quick private chat with mediator (no transfer stage). Maybe variations with anonymous/non-anonymous profiles or simple/fancy transfer? We should do the top 3-4 things people are creating.

PR #560, opened 2025-07-01
Add StockInfo stage.

rasmi:
Add StockInfo stage.

This stage can be used to present information on one or more stocks.

The stage is configured with:

- Data parsed from CSV string.
- Two default info cards (Best Year Performance and Worst Year Performance -- computed from data)
- Optional info cards (e.g. dividends, P/E ratio, etc. -- entered manually)
- A line chart generated from the data
- Optionally, show relative growth of a $1,000 investment rather than stock ticker prices.
- Optionally, require participant to view all stocks before proceeding.

Screenshot 2025-06-30 at 10 05 28 PM

Screenshot 2025-06-30 at 10 05 55 PM

Screenshot 2025-06-30 at 10 06 26 PM

Screenshot 2025-06-30 at 10 06 08 PM


frontend/src/components/stages/stockinfo_participant_view.ts
```diff
@@ -0,0 +1,233 @@
...
+          ${this.renderInfoCards(currentStock)}
+          ${this.renderMainContent(currentStock)}
+        </div>
+
+        ${this.renderStockNavigation()}
+
+        <stage-footer
+          .stage=${this.stage}
+          .disabled=${!isComplete}
+          .onNextClick=${this.saveAndProgress}
```
vivtsai, line 64: FYI, this is the same as the default `onNextClick`, so it doesn't need to be passed in

PR #576, opened 2025-07-09
Fix presence detection for multiple windows/tabs/devices

### LINKED ISSUE #575: Users shown as offline when they're actually online
It was reported that sometimes users were shown as offline, even when actively sending chat messages.

For details, see #537

raindrift:
Fixes #575 

Each tab/window maintains its own separate websocket connection. We were storing a single rtdb path for a user. If a user opened two windows, then closed one, the onDisconnect handler would be called, switching the user to offline even though they still had an open window.

This PR fixes that, by using an rtdb path with a unique id for each connection. The trigger aggregates all documents for a given user to determine if they are online or not. There's also now a scheduled job that removes old records every 72 hours.

I don't know if this fixes every issue with presence detection, but it definitely fixes _an_ issue.

This PR will have a small conflict with #572, since that one moves some of the code that temporarily disables presence detection. So if merging both of them, make sure to update #572 to remove that code (there won't necessarily be a merge conflict).

- [X] Tests pass
- [X] Appropriate changes to documentation are included in the PR

### DIFF
--- a/functions/src/presence.triggers.test.ts
+++ b/functions/src/presence.triggers.test.ts
@@ -44,13 +44,8 @@ describe('mirrorPresenceToFirestore', () => {
     child: jest.Mock<{set: jest.Mock}>;
   }
 
-  interface MockSnapshot {
-    ref: {parent: MockParent};
-    val: () => {connected: boolean; last_changed: number};
-  }
-
-  let beforeSnap: MockSnapshot;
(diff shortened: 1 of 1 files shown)


vivtsai:
Thanks (and apologies for the very late review)! Since there are merge conflicts, will close this PR and transfer the logic to a fresh PR that doesn't conflict with main.

PR #590, opened 2025-07-16
Add retry and fallback logic to chip assistance

KeHang-Zhu:
1. Change the ProposedTime Variable, Only update LLM-related fields if response was successful
2. Add the new Error mode to [utils/src/stages/chip_stage.ts](https://github.com/PAIR-code/deliberate-lab/pull/590/files#diff-91e877b3324e78c1e3dccc27dd4126f8a69bb1329b0b2b25bde83f448857c3b7)
3. In [chip_participant_view.ts](https://github.com/PAIR-code/deliberate-lab/pull/590/files#diff-243102fee910e9b77581ed63fe61e78dd33c318c11effbae8e09ee282bf0ff39), added the error mode and check if the LLM query response is not valid, render the error mode

- [x] Tests pass

### DIFF
--- a/functions/src/stages/chip.utils.ts
+++ b/functions/src/stages/chip.utils.ts
@@ -10,6 +10,7 @@ import {
   ChipTransactionStatus,
   ExperimenterData,
   ParticipantProfile,
+  ParticipantProfileExtended,
   ParticipantStatus,
   convertChipLogToPromptFormat,
   createChipInfoLogEntry,
@@ -35,6 +36,7 @@ import {
   CHIP_OFFER_ASSISTANCE_STRUCTURED_OUTPUT_CONFIG,
   CHIP_RESPONSE_ASSISTANCE_COACH_STRUCTURED_OUTPUT_CONFIG,
   CHIP_RESPONSE_ASSISTANCE_ADVISOR_STRUCTURED_OUTPUT_CONFIG,
+  ModelResponse,
 } from '@deliberation-lab/utils';
 
 import {processModelResponse} from '../agent.utils';
@@ -262,6 +264,10 @@ export async function updateChipTurn(
       return false;
     }
 
+    if (!currentTurn) {
+      return false;
+    }
+
     const currentTransaction =
       publicStage.participantOfferMap[currentRound][currentTurn];
 
@@ -342,14 +348,16 @@ export async function updateChipTurn(
         );
       }
       // Write new turn entry
-      transaction.set(
-        logCollection.doc(),
-        createChipTurnLogEntry(
-          newData.currentRound,
-          newData.currentTurn,
-          timestamp,
-        ),
-      );
+      if (newData.currentTurn) {
+        transaction.set(
+          logCollection.doc(),
+          createChipTurnLogEntry(
+            newData.currentRound,
+            newData.currentTurn,
+            timestamp,
+          ),
+        );
+      }
     }
 
     // Update public stage data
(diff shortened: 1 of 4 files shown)


vivtsai (commented):
Are you still planning to show the participant an error or other default message if none of the retries work?

functions/src/stages/chip.utils.ts
```diff
@@ -441,6 +474,62 @@ export async function addChipResponseToPublicData(
...
+      prompt,
+      modelSettings,
+      modelGenerationConfig,
+      structuredOutputConfig
+    );
+
+    lastResponse = response;
+
+    const parsed = parseResponseFn(response);
+    if (parsed.success) return parsed;
```
mkbehr, line 513: Does this branch include #510? If so, there are some error codes from ModelResponseStatus that we probably don't want to retry on, at least QUOTA_ERROR and CONFIG_ERROR.

vivtsai: Yes, good point! Since this is just for the chip pilot I think not a huge deal but @KeHang-Zhu worth considering

mkbehr:
Not quite in the scope of this PR, but I had a look at the structured output config in utils/src/stages/chip_stage.prompts.ts, and it looks like the config isn't enabling sampler constraints. If you set `type: StructuredOutputType.JSON_SCHEMA` in the `StructuredOutputConfig`, that will constrain the sampler to only output JSON matching the schema, so that should reduce the need for retries.

The Gemini API has support for specifying enum values, so we could use that to e.g. make sure the model can only output valid chip types for the chip type fields, and it doesn't use the wrong capitalization or anything. I haven't implemented that on the Deliberate Lab side, but it shouldn't be much work. Let me know if you're seeing a lot of errors of that kind after enabling JSON_SCHEMA and it'd be helpful for me to add that support.

functions/src/stages/chip.utils.ts
```diff
@@ -579,17 +587,79 @@ export async function getChipOfferAssistance(
...
+          '', // No agent public ID
+          '', // No description
+          experimenterData.apiKeys,
+          prompt,
+          modelSettings,
+          modelGenerationConfig,
+          structuredOutputConfig,
+        );
+        
+        if (response.status === 'ok') {
```
vivtsai, line 630: Use `ModelResponseStatus.OK`

functions/src/stages/chip.utils.ts
```diff
@@ -758,11 +795,71 @@ export async function getChipResponseAssistance(
     includeReasoning: true,
   });
 
+  // Helper function to call model with retries
```
vivtsai, line 798: This looks like a repeat of what you have before, just inside getChipResponseAssistance instead of getChipOfferAssistance. What do you think about moving it out of the function so that you can call the same code? (If you want to just edit both places for now that's also fine, just maybe add a comment to refactor/consolidate later and don't forget to edit in both places)

functions/src/stages/chip.utils.ts
```diff
@@ -579,17 +587,79 @@ export async function getChipOfferAssistance(
...
+        if (attempt < maxRetries) {
+          const previousText = response.text ?? '[No Text Returned]';
+          const parseErrorMessage = response.errorMessage ?? '[Unknown parse error]';
+
+          prompt =
+            basePrompt +
+            `\n\nYour previous response is:\n\`\`\`\n${previousText}\n\`\`\`\n\nParse error: ${parseErrorMessage}\n\nPlease try again.`;
+          
+          // Wait before retrying (exponential backoff)
+          await new Promise(resolve => setTimeout(resolve, Math.pow(2, attempt) * 1000));
```
vivtsai, line 647: Could you leave a note at the top of the function that retries will happen with exponential backoff? (Probably the most robust thing would be to have a boolean parameter for whether or not the retry should happen linearly or like this, but I think as long as it's clear at the top of the function should be fine)

vivtsai (approved):

PR #615, opened 2025-07-28
Fix payouts

cjqian:


vivtsai:
Changes have been cherrypicked into `chip-pilot`.

PR #620, opened 2025-07-30
undefined Chip, set default to {} for chip-pilot

KeHang-Zhu:
1. ensure chipValue is set to default {} if undefined
2. use ?. to avoid crashing on undefined values
3. Convert addchip to number before processing in the data export function

- [x] Tests pass


vivtsai:
We discussed offline and fixed the issue via the following commits:
- df199dee3681686050f9aad9a44a011fd5ca7795
- 758f9103f0a20ce799c72f04a249905f162813cf

PR #628, opened 2025-08-05
Add shuffling to PromptItemGroups.

rasmi:
Follow-up to #622 -- adds the ability to shuffle `PromptItemGroup`s based on experiment ID, cohort ID, participant ID, or a custom seed string.

(Need to rebase after #622 is merged)

### DIFF
--- a/utils/src/utils/random.utils.ts
+++ b/utils/src/utils/random.utils.ts
@@ -75,7 +75,7 @@ export const choices = <T>(array: readonly T[], n: number): T[] => {
 /** Shuffles an array using a string seed for consistent results. */
 export const shuffleWithSeed = <T>(
   array: readonly T[],
-  seedString: string,
+  seedString: string = '',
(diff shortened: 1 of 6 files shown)


utils/src/utils/random.utils.ts
```diff
@@ -1,5 +1,12 @@
 /** Random utilities that support seeding. */
 
+/** Configuration for shuffling items with different seed strategies */
+export interface ShuffleConfig {
+  shuffle: boolean;
+  seed: 'experiment' | 'cohort' | 'participant' | 'custom';
```
vivtsai, line 6: Nit: Use enum for the seed options?

vivtsai (approved):

utils/src/utils/random.utils.ts
rasmi: Done!

PR #629, opened 2025-08-05
Fix StructuredOutputType selector in agent editor

rasmi:
For some reason, the `StructuredOutputType` selector does not update when initialized. This fixes it.

### DIFF
--- a/frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
+++ b/frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
@@ -536,8 +536,7 @@ export class EditorComponent extends MobxLitElement {
           >
             <option
               value="${StructuredOutputType.NONE}"
-              ?selected=${!config.type ||
-              config.type === StructuredOutputType.NONE}
+              ?selected=${config.type === StructuredOutputType.NONE}
             >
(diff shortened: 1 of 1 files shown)


frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
```diff
@@ -531,17 +531,26 @@ export class EditorComponent extends MobxLitElement {
...
           <select
             id="structuredOutputType"
-            .selected=${config.type}
             @change=${updateType}
             ?disabled=${!this.experimentEditor.canEditStages}
           >
-            <option value="${StructuredOutputType.NONE}">
+            <option
+              value="${StructuredOutputType.NONE}"
+              ?selected=${!config.type ||
```
vivtsai, line 539: WDYT of taking `!config.type` out (as if type isn't defined, it might be misleading to show `StructuredOutputType.NONE`)?

rasmi: Done!

rasmi: (but it still does show it by default, as it did before this PR)

PR #635, opened 2025-08-11
Run log.utils.test.ts with Firestore emulator.

mkbehr:
I had a Firestore emulator already running when I was developing locally, and the test doesn't work without one.

Configure package.json to wrap this test with a firebase command to run a temporary emulator on port 8081, and check the FIRESTORE_EMULATOR_HOST environment variable it sets. Fall back to port 8080 otherwise. This way `npm test` will work whether or not an emulator is running locally, and `npx jest` will work if an emulator is running locally.

- [x] Tests pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,6 +42,10 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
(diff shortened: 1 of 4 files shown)


mkbehr:
Moving to draft since the CI doesn't like this.

PR #648, opened 2025-08-13
Add Conditional display of questions.

rasmi:
TODO:
[ ] Make sure survey_per_participant is WAI.
[ ] Add chained/grouped conditions in editor (right now you just get one condition).


frontend/src/shared/stage.utils.ts
```diff
@@ -51,6 +63,138 @@ export function isSurveyAnswerComplete(answer: SurveyAnswer | undefined) {
   return true;
 }
 
+/** Returns only questions that should be visible based on their conditions. */
```
vivtsai, line 66: Can we move these to utils/src/stages/survey_stage.ts or something like utils/src/stages/survey_stage.utils.ts? We're trying to migrate all helper functions out of this directory and into the shared utils/

rasmi:
Marked as alpha at @vivtsai's request!

PR #650, opened 2025-08-14
Collect behavioral data for bot detection

### LINKED ISSUE #649: Need better tools for detecting bot/LLM users from Prolific
Prolific has a lot of bot users, and people who use LLMs to formulate their replies to surveys. Here's a nice paper on it: https://arxiv.org/abs/2508.01390

This isn't specific to prolific. Rather, it's likely endemic to all paid research platforms, and will be a challenge forever.

We would like to have some hope of identifying these users so that we have the option of excluding them from our analysis. Doing it via analysis of their text alone is difficult-to-impossible. To that end, we'd like to record behavioral data (keystroke timing, copy/paste actions, text field change actions, etc), which will help us determine if there is a human behind the keyboard, and if they're pulling their text-based responses from elsewhere.

Using this data well is also a challenge, especially since we need to account properly for people who use assistive devices, speech-to-text, etc. However, the first step is to have the data at all.

I have a draft of some code for this ready, and will make a PR shortly. This is just the accompanying issue.

raindrift:
Fixes #649

This PR adds optional logging of click, keydown, input change, copy/paste, and window focus events. Of course, it only logs events that occur within the Deliberate Lab window. The goal is to use this behavioral data during analysis to attempt to identify users who are completing the task using a bot, and users who are having an LLM compose their survey or chat responses for them.

- [X] Tests pass

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,56 +1,65 @@
-import typescriptEslint from "@typescript-eslint/eslint-plugin";
-import globals from "globals";
-import tsParser from "@typescript-eslint/parser";
-import path from "node:path";
-import { fileURLToPath } from "node:url";
-import js from "@eslint/js";
-import { FlatCompat } from "@eslint/eslintrc";
+import typescriptEslint from '@typescript-eslint/eslint-plugin';
+import globals from 'globals';
+import tsParser from '@typescript-eslint/parser';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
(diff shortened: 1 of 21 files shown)


raindrift:
This work is based on an older commit because it is the root of the fork we are using for our present experiment. If I have time I'll come back and correct the merge conflicts. But I wanted to give you the code regardless since I expect that this is an issue everyone else will run into too (if they haven't already)

PR #655, opened 2025-08-15
Support sending hardcoded initial chat messages for agents.

rasmi:
Add initialMessage to AgentChatSettings, and adds AgentChatSettings to AgentChatPromptEditor.

(depends on  #653 and #654)

___

Screenshot 2025-08-15 at 5 25 19 PM

___

Screenshot 2025-08-15 at 5 36 11 PM


functions/src/triggers/agent_participant.triggers.ts
```diff
@@ -25,6 +25,25 @@ import {app} from '../app';
...
+    const experimentId = event.params.experimentId;
+    const participantId = event.params.participantId;
+
+    // Check if participant moved to a new stage (for initial message sending)
+    const before = event.data.before.data() as ParticipantProfileExtended;
+    const after = event.data.after.data() as ParticipantProfileExtended;
+
+    if (before.currentStageId !== after.currentStageId) {
+      // Send initial messages for chat stages
+      await sendInitialChatMessages(
```
vivtsai, line 39: This eliminates the need to call for chat messages inside completeStageAsAgentParticipant, right? Should we add a comment here and remove that call?

https://github.com/PAIR-code/deliberate-lab/blob/b5b4b88b91cdbefc9544a9427dff6c4a48d9c063/functions/src/agent_participant.utils.ts#L90

functions/src/chat/chat.agent.ts
```diff
@@ -353,6 +425,65 @@ export async function sendAgentPrivateChatMessage(
...
+) {
+  const stage = await getFirestoreStage(experimentId, stageId);
+  if (!stage) return;
+
+  // Only handle chat stages
+  if (stage.kind !== StageKind.CHAT && stage.kind !== StageKind.PRIVATE_CHAT) {
+    return;
+  }
+
+  // For chat stages, trigger initial messages from all agent mediators and participants
```
vivtsai, line 443: Technically, we should only trigger initial messages if the stage is ready to start (e.g., some stages are gated by "waiting" where all participants must be on the current stage in order for the chat to begin). Can we add a quick TODO comment here to account for that?

functions/src/chat/chat.agent.ts
```diff
@@ -65,7 +65,79 @@ export async function createAgentChatMessageFromPrompt(
...
   // Stage (in order to determin stage kind)
   const stage = await getFirestoreStage(experimentId, stageId);
+  if (!stage) return false;
+
+  // Check if this is an initial message request (empty triggerChatId)
+  if (triggerChatId === '') {
+    // Handle initial message
+    const initialMessage = promptConfig.chatSettings?.initialMessage;
+    if (!initialMessage || initialMessage.trim() === '') {
+      return false; // No initial message configured
```
vivtsai, line 75: If no initial message is configured, it should default to querying the API for an initial message.

vivtsai (changes requested):

functions/src/chat/chat.agent.ts
rasmi: Done!

PR #668, opened 2025-08-22
Use structured API for private chats

rasmi:
For private chats, this PR uses each API's native array-of-messages style rather than sending a single string prompt for the request. Depends on #667.

### DIFF
--- a/functions/src/agent.utils.ts
+++ b/functions/src/agent.utils.ts
@@ -1,20 +1,14 @@
 import {Timestamp} from 'firebase-admin/firestore';
 import {
   AgentModelSettings,
-  AgentParticipantPromptConfig,
-  AgentPersonaConfig,
   APIKeyConfig,
   ApiKeyType,
   ModelGenerationConfig,
+  ModelResponse,
   ModelResponseStatus,
-  ParticipantProfileExtended,
-  ParticipantStatus,
-  StageConfig,
-  StageKind,
   StructuredOutputConfig,
   UserProfile,
   createModelLogEntry,
-  makeStructuredOutputPrompt,
 } from '@deliberation-lab/utils';
 
 import {getGeminiAPIResponse} from './api/gemini.api';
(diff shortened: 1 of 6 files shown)


functions/src/agent.utils.ts
```diff
@@ -35,11 +29,22 @@ export async function processModelResponse(
   privateId: string,
   description: string,
   apiKeyConfig: APIKeyConfig,
-  prompt: string,
+  prompt: string | Array<{role: string; content: string; name?: string}>,
   modelSettings: AgentModelSettings,
   generationConfig: ModelGenerationConfig,
   structuredOutputConfig?: StructuredOutputConfig,
 ): Promise<ModelResponse> {
+  // Convert prompt to string for logging
```
vivtsai, line 37: I think we should ideally indicate in logging that the prompt sent in was an array—maybe we can adjust logging in a follow-up PR?

functions/src/api/openai.api.ts
```diff
@@ -116,11 +137,14 @@ export async function callOpenAIChatCompletion(
...
 
+  // Convert prompt to messages format if not already
+  const messages = convertToOpenAIFormat(prompt);
+
   let response;
   try {
     response = await client.chat.completions.create({
       model: modelName,
-      messages: [{role: 'user', content: prompt}],
+      messages: messages,
```
vivtsai, line 147: Nit: you could just write `messages,` here (since the variable name is the same as the field)

functions/src/chat/chat.agent.ts
```diff
@@ -263,7 +268,28 @@ export async function getAgentChatMessage(
...
+  const isPrivateChat = stage.kind === StageKind.PRIVATE_CHAT;
+  const useMessageFormat = shouldUseMessageFormat(
+    isPrivateChat,
+    true, // allowMessageFormat, always true for now.
+    participantIds.length,
+  );
+
+  // Prepare prompt - either message-based or traditional string
+  let prompt: string | ConversationMessage[];
+  if (useMessageFormat && isPrivateChat) {
```
vivtsai, line 281: Just a heads up that `useMessageFormat` checks for `isPrivateChat` so this is redundant right now—not sure if that's intentional

functions/src/chat/chat.agent.ts
```diff
@@ -263,7 +268,28 @@ export async function getAgentChatMessage(
     promptConfig,
   );
 
-  // Get response
+  // Check if we should use message-based format (private chat with one participant)
+  const isPrivateChat = stage.kind === StageKind.PRIVATE_CHAT;
+  const useMessageFormat = shouldUseMessageFormat(
+    isPrivateChat,
+    true, // allowMessageFormat, always true for now.
```
vivtsai, line 275: I think we ideally want to have the UI editor show two prompt options: the legacy structured prompt and this new "message format" (as technically the "prompt preview" will be inaccurate for the latter / the experimenter won't have visual confirmation that this message format will be sent to the API). Let me open an issue for better UI/logging clarity around the new message format case

vivtsai: (And then this boolean param can be whatever the user indicated in the frontend maybe?)

vivtsai (approved):

PR #683, opened 2025-08-27
Added resource-allocation consensus environments for development / prototyping. 

Aa-g-i:
For playtesting / pilot studies on resource-allocation consensus. It's worth noting that the Mediators have yet to be implemented here.


Aa-g-i:
Adding new PR after syncing / pulling.

PR #688, opened 2025-09-04
Fixes for Consensus Pilot Round2

Aa-g-i:
This does not resolve all of the feature requests from the first pilot, but it's a marked improvement, and represents a placeholder while I fix some small issues with the proper "sum to 100%" asset allocation logic for N charities...

### DIFF
--- a/frontend/src/shared/templates/charity_allocations.ts
+++ b/frontend/src/shared/templates/charity_allocations.ts
@@ -11,9 +11,7 @@ import {
   createTextSurveyQuestion,
   createScaleSurveyQuestion,
   createStageTextConfig,
-  AgentMediatorTemplate,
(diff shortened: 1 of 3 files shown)


frontend/src/shared/templates/charity_allocations.ts
```diff
@@ -1,39 +1,22 @@
 import {
-  createAgentChatSettings,
-  createAgentMediatorPersonaConfig,
-  createAgentParticipantPersonaConfig,
-  createChatPromptConfig,
+  // ⛔️ REMOVED: createMultiAssetAllocationStage,
```
vivtsai, line 2: Please don't leave comments like this (as they're not useful for the repo outside of diff review)

PR #692, opened 2025-09-05
Charity Surveys pilot round2 update

Aa-g-i:
Updated:

Order of post-discussion allocation vote and mediator survey

Merged with Vivian's multi-asset allocation feature.

Removed sociological survey questions

added positive / negative questions w.r.t mediator

Likert scale for all survey questions.

TESTED on local deployment.###


vivtsai:
Note that query selector logic (and optionally incorrectly used stages in template) should be addressed as per https://github.com/PAIR-code/deliberate-lab/pull/684 comments before merging debate-environments into main.

PR #694, opened 2025-09-12
Use (hardcoded) structured prompt for agent participant profile stage.

### LINKED ISSUE #630: Migrate agent participant profile prompt to structured prompt

mkbehr:
Fixes #630.

No support for configuring the prompt. In my manual testing, this is enough for agents to get past the profile stage, and they'll use any prompt config to set their profile if set. (Otherwise, they tend to see that e.g. their ID is orange-camel-[string] and name themselves Orange Camel with a camel emoji.)

- [x] Tests pass
- [ ] Appropriate changes to documentation are included in the PR (n/a)

### DIFF
--- a/functions/src/stages/profile.utils.test.ts
+++ b/functions/src/stages/profile.utils.test.ts
@@ -1,4 +1,3 @@
-
 import {
   AgentConfig,
   ModelResponseStatus,


cjqian (approved):
I manually tested and it looks good! Thanks for the clean code and tests. 

One small note (no action needed just to flag): The agents can pick whichever emoji they'd like. This is fun, but may be a giveaway if they have different icon sets from humans.

functions/src/stages/profile.utils.ts
```diff
@@ -57,35 +59,60 @@ export async function completeProfile(
...
+    stageConfig.id,
+    participant,
+    agentConfig,
+    promptConfig,
+  );
+
+  const response = await processModelResponse(
+    experimentId,
+    participant.currentCohortId,
+    /*participantId=*/ participant.privateId,
```
cjqian, line 90: Nit: cleanup?

functions/src/stages/profile.utils.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      );
+      const processModelResponseSpy = jest.spyOn(
+        agentUtils,
+        'processModelResponse',
+      );
+
+      const participant = {...defaultParticipant};
+      const stageConfig = {
+        ...defaultStageConfig,
+        profileType: ProfileType.ANONYMOUS_ANIMAL,
```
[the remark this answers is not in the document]
mkbehr: That's [here](https://github.com/PAIR-code/deliberate-lab/blob/ed9efae9e594ecec84c09dbe01de6935a181e545/functions/src/stages/profile.utils.ts#L53) (unchanged in this PR). It checks for ANONYMOUS_ANIMAL and ANONYMOUS_PARTICIPANT, and the [only three types today](https://github.com/PAIR-code/deliberate-lab/blob/ed9efae9e594ecec84c09dbe01de6935a181e545/utils/src/stages/profile_stage.ts#L15) are those and DEFAULT.

mkbehr:
> One small note (no action needed just to flag): The agents can pick whichever emoji they'd like. This is fun, but may be a giveaway if they have different icon sets from humans.

Good point. We can leave as is for now, but if we want to limit this later we could change the emoji output type to an enum.

functions/src/stages/profile.utils.ts
mkbehr: The `/*participantId=*/` is a [parameter-name comment](https://google.github.io/styleguide/tsguide.html#comments-when-calling-a-function) to make it clear what parameter this is going to.

PR #701, opened 2025-09-25
Simplify agent mediator stage-specific prompts

### LINKED ISSUE #700: Simplify agent mediator state-specific prompts
Make it clearer to users that only some of the stages are eligible for prompts, and collapse the stages that don't need editing to allow users to find the stage they need faster.

nadah09:
Fixes #700

Collapses the inactive stage descriptions to one line to avoid taking up unnecessary vertical space, and grays out stages that cannot have prompts to allow users to find the relevant stages sooner.

image

### DIFF
--- a/frontend/src/components/experiment_builder/agent_chat_prompt_editor.scss
+++ b/frontend/src/components/experiment_builder/agent_chat_prompt_editor.scss
@@ -14,7 +14,7 @@
 
 .inactive-stage {
   @include common.flex-row;
-  gap: 8px;
+  gap: common.$spacing-medium;
   align-items: baseline;
 }
 


frontend/src/components/experiment_builder/agent_chat_prompt_editor.scss
```diff
@@ -12,6 +12,16 @@
   padding: common.$main-content-padding;
 }
 
+.inactive-stage {
+  @include common.flex-row;
+  gap: 8px;
```
vivtsai, line 17: nit: use `common.$spacing-medium` for `8px`

vivtsai (approved):

frontend/src/components/experiment_builder/agent_chat_prompt_editor.scss
nadah09: done!

PR #705, opened 2025-10-02
Add model reasoning fields to agent mediator frontend

### LINKED ISSUE #601: Add Gemini 2.5 model reasoning fields to frontend UI
The fields are `reasoningBudget` and `includeReasoning`: https://github.com/PAIR-code/deliberate-lab/blob/main/utils/src/agent.ts#L62.

We probably want to have a note that these are only used for Gemini 2.5 model.

The UI field should go in https://github.com/PAIR-code/deliberate-lab/blob/main/frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts.

nadah09:
Fixes #601

Adds includeReasoning checkbox and reasoning budget field to agent mediator advanced settings.

image

### DIFF
--- a/frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
+++ b/frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
@@ -184,6 +184,8 @@ export class EditorComponent extends MobxLitElement {
         <div class="divider"></div>
         ${this.renderAgentChatSettings(this.agent, promptConfig)}
         <div class="divider"></div>
+        ${this.renderModelReasoningParameters(this.agent, promptConfig)}
+        <div class="divider"></div>
         ${this.renderAgentSamplingParameters(this.agent, promptConfig)}
         <div class="divider"></div>
         ${this.renderAgentCustomRequestBodyFields(this.agent, promptConfig)}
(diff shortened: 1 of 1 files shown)


frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
```diff
@@ -444,6 +446,69 @@ export class EditorComponent extends MobxLitElement {
...
+            ?disabled=${!this.experimentEditor.isCreator}
+            @click=${updateIncludeReasoning}
+          >
+          </md-checkbox>
+          <div>Include model reasoning</div>
+        </div>
+
+        <div class="field">
+          <label for="reasoning-budget">Reasoning Budget</label>
+          <div class="description">Reasoning budget for Gemini 2.5</div>
```
nadah09, line 498: Wasn't sure what the best text description for this is, let me know what would be more useful to change it to

frontend/src/components/experiment_builder/agent_chat_prompt_editor.ts
```diff
@@ -444,6 +446,69 @@ export class EditorComponent extends MobxLitElement {
...
+          <div class="number-input">
+            <input
+              .disabled=${!this.experimentEditor.isCreator}
+              type="number"
```
nadah09, line 502: What min and max values should I cap this to?

cjqian:
LGTM. Stylistic comment - given that certain parameters are only supported for certain models (e.g. OpenAI, Gemini 2.5); can we consider "disabling" fields / greying out advanced parameters when the selected model is not of the current type?

A coarse Gemini / OpenAI / other check should be fine to start - being more fine grained than that (e.g. different types of Gemini) might be trickier.

PR #708, opened 2025-10-15
Add initial consensus templates to experiment builder

vivtsai:
Add new "resource allocation" and "charity allocation" experiment templates, both of which require user input (specified topic and specified inclusion/exclusion of stages, respectively) when being configured in the experiment builder "load template" dialog.

This PR should be squashed and merged.

PR #710, opened 2025-10-17
Clean up template for Hubrec consensus - post HubREC

Aa-g-i:
Implemented feedback from HuBREC and Vivian, specifically unhelpful comments, and lack of linting.

### DIFF
--- a/frontend/src/shared/templates/charity_allocations.ts
+++ b/frontend/src/shared/templates/charity_allocations.ts
@@ -79,8 +79,7 @@ const CONSENSUS_TOS_STAGE = createTOSStage({
   tosLines: [
     'Thank you for your interest in this research. If you choose to participate, you will be asked to participate in debates about resource allocation, which have real-world consequences in the form of disbursing real funds to real charities, based on your / your teams actions.',
     '**Compensation & Impact**',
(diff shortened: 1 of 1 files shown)


cjqian:
Pushed a few small changes. Some feedback:

* We need to prime mediators before "Initial mediator survey." E.g. a new info stage explaining that a mediator might intervene in the discussions, maybe an example of what that looks like. 
* Similarly, images of the allocation screen may helpful folks understand. I'd split 4. into two parts (one on how allocations / consensus work, another on impact)
* Beginning info text stages: "You are now beginning Round 1" is not enough context. Consider adding like "You will not have an AI intervention in this stage" so that expectations are set. This is a good spot to also introduce the three charities. (Tip: You can add images, here and above, through image markdown references. I use imgur to host)
* Throw a timer on the chat discussions and describe this to participants (the "Disable conversation after a fixed amount of time" feature)
* It's important that each conversation hasthe same post-conversation survey discussion questions, with a comparative mediator at the end of the experiment (stage 25). In the consensus doc, let's iterate on the survey questions (throw them in there with what each question aims to measure)!

vivtsai:
Hi @cjqian @Aa-g-i - going to merge this into `debate-environments` branch now so we can compare the total diff of PR #708 (which we plan to squash and merge) with main.

@Aa-g-i heads up that the formatting test here doesn't seem to have passed, so if you could fix that'd be great.

PR #711, opened 2025-10-17
Add typecheck commands and fix typecheck errors in functions.

mkbehr:
Add `npm run typecheck` commands to `utils` and `functions`, and begin fixing the typecheck errors in `functions`.

I couldn't quickly find a way to add a typecheck command to `frontend`'s webpack setup, but webpack performs typechecking on build anyway, so it's not as necessary. `utils` is also clean, which I think is because it uses tsup. `functions` is the main culprit for typechecking errors.

This reduces the number of typechecking errors in `functions` from 473 to 252.

This also creates a new "Developing" doc page for the linting commands and the new typecheck commands.

- [x] Tests pass

### DIFF
--- a/functions/src/agent.endpoints.ts
+++ b/functions/src/agent.endpoints.ts
@@ -2,12 +2,8 @@ import {
   ModelResponseStatus,
   createModelGenerationConfig,
 } from '@deliberation-lab/utils';
-import {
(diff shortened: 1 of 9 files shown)


vivtsai:
Thank you for working on this!!

Issue #713, opened 2025-10-20
Firebase authentication: Allow e-mail checks to be case-insensitive


cjqian:
If you sign up with FooBar@gmail.com, foobar@gmail.com should be able to use the platform

PR #720, opened 2025-10-21
Make email logic case insensitive for verification and authentication. 

aldenhallak:
## Description
Make email logic case insensitive for verification and authnetication. 

---

## Verification
- [ ] All tests pass (if applicable)

---

## Related issues
This PR fixes: #713 

---
[Proof that it still works.webm](https://github.com/user-attachments/assets/3917f33d-cd6b-4aa0-ac84-f7d402cb3f7e)


firestore/firestore.rules
```diff
@@ -33,7 +33,7 @@ service cloud.firestore {
 
     // Function: checks if creator of experiment
     function isExperimentCreator(experimentId) {
-      return get(/databases/$(database)/documents/experiments/$(experimentId)).data.metadata.creator == request.auth.token.email || get(/databases/$(database)/documents/experiments/$(experimentId)).data.metadata.creator == request.auth.uid;
+      return get(/databases/$(database)/documents/experiments/$(experimentId)).data.metadata.creator == request.auth.token.email.lower() || get(/databases/$(database)/documents/experiments/$(experimentId)).data.metadata.creator == request.auth.uid;
```
vivtsai: Actually, we could avoid having to migrate the creator field (for existing experiments in the database where it may have been stored with capitals) by having `creator.lower()` here?

aldenhallak: I went ahead and did this (in a bit more comprehensive way - check the latest commit) but we can't really modify these two functions in the same capacity:

```
    // Function: check if admin
    function isAdmin() {
      return get(/databases/$(database)/documents/allowlist/$(request.auth.token.email.lower())).data.get("isAdmin", null) != null;
    }

    // Function: checks if experimenter
    function isExperimenter() {
      return request.auth.uid != null && exists(/databases/$(database)/documents/allowlist/$(request.auth.token.email.lower()));
    }
   
```
    
Firebase doc paths are case sensitive, so we need to make sure that all the values are lowercase on entry.

functions/src/scripts/normalize-allowlist-emails.ts
vivtsai: This can be removed since we're using the endpoint right?

aldenhallak: Good point... removed!

PR #738, opened 2025-10-28
Add checkbox so we can disable the automatic adding of a mediator in a cohort.

aldenhallak:
## Description
Add "Automatically add mediator to every cohort" checkbox to UI (checked by default), and then (if the box is not checked) allow people to add mediators to cohort in the dashboard.

We add the createMediator callable so experimenters can opt out of auto-attaching every mediator persona to new cohorts but still attach them later from the dashboard.

Before: Cohort creation used createMediatorsForCohort to auto-create mediators for every persona where isDefaultAddToCohort was true, but there was no pathway to add one afterwards if that flag was false.

We introduce createMediator cloud function/endpoints createMediatorCallable and shared createMediatorForCohortFromPersona helper. This exposes backend functionality to build a mediator profile from a persona (fetch prompts, set IDs, write to Firestore) on demand.

Without it, the new checkbox would orphan personas with no way back into a cohort, breaking the user flow.

---

## Verification

---
[Adding mediators](https://github.com/user-attachments/assets/fa89fae1-8e85-4405-a0c0-8b0841844507)

## Related issues
This PR fixes: #743

---

### DIFF
--- a/frontend/src/components/experiment_dashboard/cohort_editor.ts
+++ b/frontend/src/components/experiment_dashboard/cohort_editor.ts
@@ -8,6 +8,7 @@ import './cohort_summary';
 import './participant_summary';
 import './agent_participant_configuration_dialog';
 import './agent_mediator_add_dialog';
+import {renderMediatorStatusChip} from './mediator_status';
 
 import {MobxLitElement} from '@adobe/lit-mobx';
 import {CSSResultGroup, html, nothing, TemplateResult} from 'lit';
@@ -31,7 +32,6 @@ import {
   ParticipantProfileExtended,
   ParticipantStatus,
   StageKind,
-  getAgentStatusDisplayText,
 } from '@deliberation-lab/utils';
 
 import {styles} from './cohort_editor.scss';
@@ -207,14 +207,7 @@ export class Component extends MobxLitElement {
 
     const renderMediator = (mediator: MediatorProfileExtended) => {
       const renderStatus = () => {
-        if (!mediator.agentConfig) {
-          return nothing;
-        }
-        return html`
-          <div class="chip secondary">
-            🤖 ${getAgentStatusDisplayText(mediator.currentStatus)}
-          </div>
-        `;
+        return renderMediatorStatusChip(mediator);
       };
 
       const toggleStatus = async () => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/experiment_builder/agent_persona_editor.ts
```diff
@@ -233,6 +235,30 @@ export class AgentPersonaEditorComponent extends MobxLitElement {
...
+    return html`
+      <div class="checkbox-wrapper">
+        <md-checkbox
+          touch-target="wrapper"
+          ?checked=${!agent.isDefaultAddToCohort}
+          ?disabled=${!this.experimentEditor.canEditStages}
+          @change=${togglePreference}
+        >
+        </md-checkbox>
+        <div>Don't automatically add mediator to cohort</div>
```
vivtsai: Thanks! This is my fault for wording the issue this way, but: could we actually have this be "Automatically add this mediator to every cohort" and have it checked by default? (Guessing that might be easier to understand in the UI.)

vivtsai:
Could we also add a quick line or two to the docs about being able to specify if mediator is automatically added/not and if not, that you can add from the dashboard? Maybe [run-experiment.md](https://pair-code.github.io/deliberate-lab/researchers/run-experiment) or [add-mediator.md](https://pair-code.github.io/deliberate-lab/researchers/add-mediator)

frontend/src/components/experiment_dashboard/cohort_editor.ts
```diff
@@ -207,14 +207,7 @@ export class Component extends MobxLitElement {
...
       const renderStatus = () => {
-        if (!mediator.agentConfig) {
-          return nothing;
-        }
-        return html`
-          <div class="chip secondary">
-            🤖 ${getAgentStatusDisplayText(mediator.currentStatus)}
-          </div>
-        `;
+        return renderMediatorStatusChip(mediator);
```
vivtsai: Not convinced this should be moved to a different file... WDYT about keeping it here since the status isn't used elsewhere?

aldenhallak: It's used in the editor and the summary!

vivtsai: Oh duh! Thought I might be missing something.

PR #746, opened 2025-10-30
Add variables to experiments

rasmi:
[TO ADD DOCS]


vivtsai:
Discussed offline - going with slightly different approach but can borrow some elements from this design.

Issue #764, opened 2025-10-31
Add warning dialog if experimenter tries to transfer participant with duplicate name into same cohort


vivtsai:
@nadah09 when we set up auto-cohort transfer, we should also check against duplicate names for participants.

PR #770, opened 2025-11-03
add confirmation dialog when transferring a participant to a cohort that already contains a participant with the same name

aldenhallak:
## Description
Now we can't transfer a participant when they have the same name. In order to test this, I had to add debug features, which I have not committed to this branch.

[cast (1).webm](https://github.com/user-attachments/assets/56601546-dd4a-48ba-ada9-11b40c5aca51)

---

## Verification

---

## Related issues
This PR fixes: #764 

---


vivtsai:
It looks like this needs another Prettier run but otherwise change looks good!

Although... do we want to also check for anonymous base name duplicates, e.g., "Bat" vs. "Bat 2"? can be compared using `AnonymousProfileMetadata` if `name` is the same and also being used in the `name` field (or by parsing out the "animal" part of the public ID). In that case could be nice to have a utils function that compares profiles with a param for whether/not to check for base name duplicates. I'm fine if we table that for later though

vivtsai (approved):

PR #771, opened 2025-11-03
Cleaned up charity_allocation.ts with post-pilot feedback

Aa-g-i:
Per offline discussions, we can't dynamically shift the stages on a per-cohort basis without significant refactors. This represents a compromise, where we made it as straightforward as possible to spin up 3! different experiments representing all 3! combinations of [No Mediator, Habermas Mediator, Dynamic Mediator].

Changed max conversation time to a minimum to enforce longer conversations on ideal allocations. 
Cleaned up wording to keep mediator from getting fixated on optimality.
TODO: check Consensus score computations. 

TESTED=initializing experiments and confirming each setting properly includes mediators / mediator surveys and are reflected properly in the final mediator preference survey.

### DIFF
--- a/docs/code-of-conduct.md
+++ b/docs/code-of-conduct.md
@@ -69,8 +69,8 @@ dispute. If you are unable to resolve the matter for any reason, or if the
 behavior is threatening or harassing, report it. We are dedicated to providing
 an environment where participants feel welcome and safe.
 
-Reports should be directed to *[PROJECT STEWARD NAME(s) AND EMAIL(s)]*, the
-Project Steward(s) for *[PROJECT NAME]*. It is the Project Steward’s duty to
+Reports should be directed to llm-mediation-core@google.com, the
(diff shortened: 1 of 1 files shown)


frontend/src/shared/templates/charity_allocations.ts
```diff
@@ -145,7 +150,7 @@ const CHARITY_DATA: CharityInfo[] = [
     key: 'clean_ocean',
     name: '🌊 Clean Ocean Action',
     link: 'https://www.charitynavigator.org/ein/222897204',
-    score: 'Not Rated', // Score not present in link text
+    score: '99%%',
```
vivtsai: Should this be `99%`?

vivtsai (approved):

vivtsai (approved):

vivtsai (approved):

PR #772, opened 2025-11-04
Enable experimenters to set string/object variables in select stage text fields

vivtsai:
## Description
This introduces MVP support for variables within experiments. These can be configured via the experiment builder as either string or object (values specified in JSON for now) variables and with either cohort-level or participant-level assignment.

[Example variable setup](https://github.com/user-attachments/assets/fe864a1e-1678-4a2d-bd0b-f83c93d4bcd4)

---

## Verification
- [x] *Screenshots or videos* included
- [x] Clear reproduction steps provided to invoke the feature - see video
- [x] All tests pass

### DIFF
--- a/docs/code-of-conduct.md
+++ b/docs/code-of-conduct.md
@@ -69,8 +69,8 @@ dispute. If you are unable to resolve the matter for any reason, or if the
 behavior is threatening or harassing, report it. We are dedicated to providing
 an environment where participants feel welcome and safe.
 
-Reports should be directed to *[PROJECT STEWARD NAME(s) AND EMAIL(s)]*, the
-Project Steward(s) for *[PROJECT NAME]*. It is the Project Steward’s duty to
+Reports should be directed to llm-mediation-core@google.com, the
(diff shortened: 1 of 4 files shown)


cjqian:
Amazing! One thing that was confusing to me: Does each "Variable group" map to 1 "Variable name"? In the examples with charities and cities, what happens if I press "Add variable" multiple times in that group? (E.g. City, City2)?

My guess looking at the UI is that there should be 1 "Name of variable group" field (not a many:1 + Add variable button per group); please let me know if this is not the case.

image

frontend/src/components/experiment_builder/variable_editor.ts
```diff
@@ -0,0 +1,239 @@
...
+            ...variableConfig.values.slice(valueIndex + 1),
+          ],
+        },
+        configIndex,
+      );
+    };
+
+    return html`
+      <pr-textarea
+        placeholder="Value"
```
cjqian: Give an example here, like "Value string or json object depending on type, e.g. {"foo":"bar"}"?

cjqian (approved):

PR #780, opened 2025-11-05
Implement v0 of improved logs dashboard with sorting and filtering. 

cjqian:
Key diffs:
- Moved from singular participant panel to experiment panel (by default, you can see all logs over all cohorts / participants anyway)
- Implement filtering and sorting
- Clean up interface (no more spillover, expandable fullscreen toggle)

https://github.com/user-attachments/assets/60a7bdd5-941e-4d9c-8882-a7cac82e0751

### DIFF
--- a/frontend/src/components/experimenter/log_dashboard.ts
+++ b/frontend/src/components/experimenter/log_dashboard.ts
@@ -8,6 +8,7 @@ import {customElement, state} from 'lit/decorators.js';
 
 import {core} from '../../core/core';
 import {ExperimentManager} from '../../services/experiment.manager';
+import {ExperimentService} from '../../services/experiment.service';
 
 import {
   convertUnifiedTimestampToDateTime,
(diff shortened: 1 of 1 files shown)


frontend/src/components/experimenter/log_dashboard.scss
```diff
@@ -0,0 +1,179 @@
...
+  background-color: var(--md-sys-color-surface);
+  color: var(--md-sys-color-on-surface);
+  padding: 20px;
+  border: 1px solid var(--md-sys-color-outline);
+  width: var(--pr-text-area-width, 80%);
+  max-height: 90vh;
+  height: auto;
+}
+
+/* Sort + Filters */
```
vivtsai: Can we take out these auto-comments?

frontend/src/shared/utils.ts
```diff
@@ -37,6 +37,15 @@ export function collectSnapshotWithId<T>(snapshot: Snapshot, idKey: keyof T) {
   return snapshot.docs.map((doc) => ({[idKey]: doc.id, ...doc.data()}) as T);
 }
 
+/** Returns the number of seconds between start time and end time. */
+export function getUnifiedDurationSeconds(
```
vivtsai: Please don't add utility functions to the frontend—we're in the process of trying to migrate everything out of here and into the `utils/` directory.

vivtsai:
FWIW I would be fine with keeping the log-dashboard component under `experiment-dashboard/` directory so that all the components of the experiment dashboard are eventually under there. But your call

PR #783, opened 2025-11-05
Use Node 22 for Cloud functions

vivtsai:
Runtime Node.js 18 was decommissioned on 2025-10-31


aldenhallak (approved):
hopefully this is fine :)

PR #787, opened 2025-11-06
Add scaffolding to default agent participant/mediator prompts

cjqian:
### Key diffs:
* Differentiation between previous stages and old stages
* Richer persona instantiation, including clear distinction between randomly-generated pseudonyms and intentionally chosen profiles
* Cleaner instructions / transcript separation and formatting in chat transcript

Examples in this gist: https://gist.github.com/cjqian/ba92c38bc0442892956d83bd486009ac

### DIFF
--- a/frontend/src/shared/templates/charity_allocations.ts
+++ b/frontend/src/shared/templates/charity_allocations.ts
@@ -1,25 +1,19 @@
 import {
+  AgentPersonaType,
   createTextPromptItem,
   createChatStage,
   createDefaultStageContextPromptItem,
   AgentMediatorTemplate,
   MediatorPromptConfig,
-  AgentPersonaType,
   createAgentMediatorPersonaConfig,
-  ChatPromptConfig,
-  ChatMediatorStructuredOutputConfig,
+  createParticipantProfileBase,
   StructuredOutputDataType,
-  StructuredOutputType,
   StructuredOutputSchema,
   createStructuredOutputConfig,
-  createAgentChatPromptConfig,
   createAgentChatSettings,
   PromptItemType,
   ProfileInfoPromptItem,
   ProfileContextPromptItem,
-  StageContextPromptItem,
-  TextPromptItem,
-  AgentChatSettings,
   DEFAULT_AGENT_MODEL_SETTINGS,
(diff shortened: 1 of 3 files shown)


functions/src/structured_prompt.utils.ts
```diff
@@ -336,34 +326,73 @@ async function processPromptItems(
...
+        // Previous stages for non-mediators
+        const ids = stageContextIds;
+        if (userProfile.type === UserType.PARTICIPANT) {
+          items.push('\n--- Previously completed stages (read only) ---');
+
+          const prevIds = ids.slice(0, -1); // all except last
+
+          for (let i = 0; i < prevIds.length; i++) {
+            const id = prevIds[i];
+            const stageNumber = i + 1; // 1-based
```
vivtsai: Does it matter that this stage number may not be accurate? E.g., if you only pass 4 of the 8 previous stages in as context. Not to mention if you only pass in one at a time when configuring, they'll all render as stage 1 here.

If the actual number is important, it could be pulled from the `experiment.stageIds` (already available in this function)

functions/src/structured_prompt.utils.ts
```diff
@@ -324,7 +310,11 @@ async function processPromptItems(
         items.push(promptItem.text);
         break;
       case PromptItemType.PROFILE_CONTEXT:
-        items.push(userProfile.agentConfig.promptContext);
+        if (userProfile.agentConfig?.promptContext) {
+          items.push(
+            `Private persona context: ${userProfile.agentConfig.promptContext}\nThis information is private to you. Use it to guide your behavior in this task. Other participants do not know these attributes unless you choose to share it.`,
```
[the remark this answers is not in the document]
cjqian: Discussed offline; will leave for participants as this can't be edited anyway, and correct for mediators.

functions/src/structured_prompt.utils.ts
cjqian: Removed stage number for now to work around this. So the separation between previous and current stages remain, but no new stage numbers are added.

cjqian:
@vivtsai PTAL.

PR #789, opened 2025-11-06
Add partial fix for resolving variables in prompt stage display

vivtsai:
This passes stage configs through variable resolution before assembling the stage context prompt item.

WARNING: This will not work as intended if the prompt has multiple participants passed in for answer rendering and the variables are set at the participant level; in this case, only the "current" user's participant variables will be shown. This will be fixed in a follow-up.

[Resolve variables in stage displays.webm](https://github.com/user-attachments/assets/ac1a3996-383e-4711-91c9-8a098df1e1ef)

- [x] Documentation added in comments


cjqian (commented):
LGTM!

PR #828, opened 2025-11-14
Add Claude API support for agents and add new charity allocation prompt

Aa-g-i:
TESTED with unittests and confirming that we can make calls via the emulated frontend...


Aa-g-i:
Per offline conversation with Crystal, forked charity_allocations.ts to charity_allocations_ootb.ts, an environment which uses a default mediator prompt across three frontier models (claude / openai / gemini)

Issue #825, opened 2025-11-13
Add live timer countdown to group chat stage


aldenhallak:
Add a live timer when there's a time limit so users can see how much time is left.

PR #843, opened 2025-11-18
Add timer in participant header that shows how much time is left.

aldenhallak:
## Description

Shows a timer in the participant header to see how much time is lesft.

[Show timer .webm](https://github.com/user-attachments/assets/1ff672dc-02ae-4537-b941-6b3ef7ef6ad3)

---

## Verification
- [ ] All tests pass (if applicable)

---

## Related issues
This PR fixes: #825

---

### DIFF
--- a/frontend/src/components/participant_view/participant_header.ts
+++ b/frontend/src/components/participant_view/participant_header.ts
@@ -176,7 +176,12 @@ export class Header extends MobxLitElement {
 
     return html`
       <div class="timer">
-        <pr-icon icon="timer" color="primary" size="large" variant="default"></pr-icon>
+        <pr-icon
+          icon="timer"
+          color="primary"
+          size="large"
+          variant="default"
+        ></pr-icon>
         ${timeString}
       </div>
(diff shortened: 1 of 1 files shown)


frontend/src/components/participant_view/participant_header.ts
```diff
@@ -2,17 +2,26 @@ import '../../pair-components/button';
 import '../../pair-components/icon_button';
 import '../../pair-components/info_popup';
 import '../../pair-components/tooltip';
+import '../../pair-components/icon'; // Added icon import
```
vivtsai: Please remove AI comments

vivtsai:
> @vivtsai -- I'm worried I don't know what other non-chat stages would have the timer like this. The wait stage has a different timer implementation, but outside of that I'm not sure.

Thanks! Right now it's just chat so no worries. By "wait stage" do you mean transfer stage? A waiting screen could have a counting-up time check (how long you've been waiting) but we don't have a limit there.

Issue #833, opened 2025-11-17
Make agent participant selection panel model-based rather than persona-based


nadah09:
Allow experimenters to choose between Gemini 2.5 Flash and Pro when making their agent participants, with no persona-based options available.

PR #864, opened 2025-11-20
Change agent participants to be model-based rather than persona-based

nadah09:
## Description
Changes agent participants to be model-based rather than persona-based. Users can select between Gemini 2.5 Flash and Pro, and add some prompt context before adding an agent to their experiment. 
image

---

## Related issues
This PR fixes: #833 
---

### DIFF
--- a/frontend/src/components/experiment_dashboard/agent_participant_configuration_dialog.ts
+++ b/frontend/src/components/experiment_dashboard/agent_participant_configuration_dialog.ts
@@ -86,8 +86,7 @@ export class AgentParticipantDialog extends MobxLitElement {
             );
             if (this.cohort && this.model) {
               this.experimentEditor.addAgentParticipant();
-              this.agentId =
-                this.experimentEditor.currentAgent?.persona.id ?? '';
+              this.agentId = ''; //Make agent ID blank for agents added from cohort panel that use default prompts
               const modelSettings = createAgentModelSettings({
                 apiType: ApiKeyType.GEMINI_API_KEY,
                 modelName: this.model,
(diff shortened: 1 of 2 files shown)


frontend/src/components/experiment_dashboard/agent_participant_configuration_dialog.ts
```diff
@@ -66,25 +74,29 @@ export class AgentParticipantDialog extends MobxLitElement {
...
+              const modelSettings = createAgentModelSettings({
+                apiType: ApiKeyType.GEMINI_API_KEY,
+                modelName: this.model,
+              });
+
               this.experimentManager.createAgentParticipant(this.cohort.id, {
                 agentId: this.agentId,
                 promptContext: this.promptContext,
-                modelSettings: agent.defaultModelSettings,
+                modelSettings: modelSettings,
```
vivtsai: Nit: You can just put `modelSettings` here since the field name and the variable are the same!

frontend/src/components/experiment_dashboard/agent_participant_configuration_dialog.ts
```diff
@@ -66,25 +74,29 @@ export class AgentParticipantDialog extends MobxLitElement {
...
               ButtonClick.AGENT_PARTICIPANT_ADD,
             );
-            if (this.cohort && this.agentId) {
-              const agent =
-                this.experimentManager.agentPersonaMap[this.agentId] ??
-                this.experimentEditor.getAgentParticipant(this.agentId)
-                  ?.persona;
+            if (this.cohort && this.model) {
+              this.experimentEditor.addAgentParticipant();
+              this.agentId =
```
vivtsai: In the old workflow, the agent ID needed to match the persona ID so that it could fetch the correct prompts (associated with the agent persona configured in the experiment editor).

Here, I think it makes sense to automatically make the ID blank (or some default value) since we want all agent participants to use "default" agent participant prompts.

To that point: What do you think about hiding the agent participant panel in the experiment editor so there's no confusion? (Could also add comments in that spot, as well as in the ExperimentEditorService, so we remember that that workflow for configuring participants is deprecated.)

Issue #868, opened 2025-11-22
Upgrade firebase-functions to v7 by Dec 31, 2025


vivtsai:
"If you have been using environment configuration with functions.config, you must migrate your existing configuration to a supported format before the end of 2025, when functions.config will be decommissioned. After December 2025, new deployments with functions.config will fail."

https://firebase.google.com/docs/functions/config-env#migrate-config

Attempted to do this in #863 but it didn't work, so temporarily downgraded to v6.

PR #869, opened 2025-11-22
Temporarily downgrade firebase-functions to v6 to avoid breaking platform

vivtsai:
This was upgraded to v7 as part of #863, but running with v7 breaks the Cloud functions endpoints and yields the following error message:

"functions.config() has been removed in firebase-functions v7. Migrate to environment parameters using the params module. Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config"

To avoid breaking the platform, use v6 for now while we identify and resolve the issue. (This PR cleans up all known uses of obsolete firebase-functions.)

Filed https://github.com/PAIR-code/deliberate-lab/issues/868 to resolve by end of 2025.

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -27,7 +27,7 @@
     "@google/genai": "^1.30.0",
     "@sinclair/typebox": "^0.34.41",
     "firebase-admin": "^13.6.0",
-    "firebase-functions": "^7.0.0",
+    "firebase-functions": "^6.0.0",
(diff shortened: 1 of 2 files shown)


aldenhallak (approved):

aldenhallak:
Interesting... functions.config() is used in an old version of firebase-tools. Upgrading firebase-tools fixes this.

https://github.com/firebase/firebase-tools/issues/8925

PR #889, opened 2025-12-03
Enable easy template validation in stages

rasmi:
(depends on #888 and #890)
Add `pr-textarea-template` component that extends `pr-textarea` with automatic Mustache variable validation. Update stage editors to use it for fields that support template variables (structured prompts, surveys, flipcards, info stages).


rasmi:
Fixed!

PR #901, opened 2025-12-10
Add ability to maximise images in chat

marksgraham:
## Description

Allow images to be maximised in chat. It was sometimes hard to make out the details of displayed images, especially those containing text.

## Verification

image
image

### DIFF
--- a/frontend/src/components/chat/chat_message.ts
+++ b/frontend/src/components/chat/chat_message.ts
@@ -48,6 +48,7 @@ export class ChatMessageComponent extends MobxLitElement {
   };
 
   private showFullscreen = (file: StoredFile) => {
+    if (!file) return;
     this.closeFullscreen();
 
     const el = document.createElement(


frontend/src/components/chat/chat_message.ts
```diff
@@ -32,6 +32,148 @@ export class ChatMessageComponent extends MobxLitElement {
...
+        cursor: pointer;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        box-shadow: 0 2px 8px rgba(0, 0, 0, 0.2);
+        transition: background 0.2s ease;
+      }
+
+      .chat-image-modal .close-button:hover {
+        background: #f0f0f0;
```
vivtsai: Use color variables for background so that this will work well with both light and dark modes

jimbojw:
Thanks Mark, I see what you're going for here. I'm still onboarding to the project, so I can't speak with perfect authority yet overall, but I have a compartmentalization recommendation.

Your proposal to attach an object to the `document.body` is fine for a quick solution. I think we can streamline it a bit.

Currently you're creating DOM elements inside the chat message's `renderModalToBody()` method and giving them inline style and behavior. Instead, I'd recommend making a new, separate `FullscreenImage` component. This object would have all the SCSS and callbacks (close button etc.) in its `fullscreen_image.scss` and `fullscreen_image.ts` respectively.

Then, in the `chat_message.ts`, you _should_ be able to create a `new FullscreenImage()` and attach that to `document.body` to get the same effect.

marksgraham:
Hi Jimbo,

Thanks for the comments - hopefully the commit addresses them, please let me know if you have any further thoughts!

PR #918, opened 2025-12-15
Fix question condition validation in survey editor.

rasmi:
Survey questions with conditions could become invalid when reordered (e.g., moving a question with a condition up before the question it depends on). This caused the question to not render in the participant view. 

Added sanitizeQuestionConditions() which automatically clears conditions that reference questions that no longer exist or come later in the survey. This runs whenever questions are moved or deleted.

This WON'T fix existing experiments that have configurations from prior to this change.

### DIFF
--- a/frontend/src/components/stages/survey_editor.ts
+++ b/frontend/src/components/stages/survey_editor.ts
@@ -12,10 +12,10 @@ import {renderConditionEditor} from '../../shared/condition_editor.utils';
 import {
   CheckSurveyQuestion,
   Condition,
-  extractConditionDependencies,
   getConditionTargetsFromStages,
   MultipleChoiceItem,
   MultipleChoiceSurveyQuestion,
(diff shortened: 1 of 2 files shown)


frontend/src/components/stages/survey_editor.ts
```diff
@@ -103,36 +104,87 @@ export class SurveyEditor extends MobxLitElement {
...
+  /**
+   * Sanitize question conditions to ensure they only reference valid targets.
+   * A condition is invalid if it references:
+   * - A question that doesn't exist in the list
+   * - A question that comes at or after the current question's position
+   *
+   * This is called automatically when updating questions to handle any
+   * reordering, deletion, or other changes that might invalidate conditions.
+   */
+  private sanitizeQuestionConditions(
```
vivtsai: Possible to move to utils?

vivtsai (changes requested):

Issue #909, opened 2025-12-12
Experiments (not just old version experiments) aren't rendering experiment preview


cjqian:
Regression where participant previews do not render:
There are console log errors related to a name; possible new feature or **variable**-related.

Experiment links for reproducing:
* https://deliberate-lab.appspot.com/#/e/02ce565c-c864-4bfe-9fe5-a920e269ce25/p/dea75a8f-f7f7-4edc-ae5b-4718e6e4dc9d 
* https://deliberate-lab.appspot.com/#/e/04307330-0e3a-41ef-a5d1-8b4194bcdf6a

jimbojw:
Does this affect all experiments, or just some?

PR #928, opened 2025-12-16
Fix variable editor schema handling for multi-value configs.

rasmi:
Fix variable editor schema handling for multi-value configs (e.g. RANDOM_PERMUTATION, BALANCED_ASSIGNMENT)

### Problem

When editing BALANCED_ASSIGNMENT variables in the experiment builder:
1. Creating an "Array of Object" as a nested property would incorrectly collapse to just "Object"
2. The editor showed "Array items: X" instead of "Type: X" for the root item type, inconsistent with loaded templates
3. BALANCED_ASSIGNMENT wasn't handled the same as RANDOM_PERMUTATION for multi-value config operations

### Issue & fix
The core issue was in `updateSchemaAtPath`: when updating a property that was an array (e.g., `charities: Array(Object({}))`), and the path ended at that property, it replaced the entire property with the new schema instead of updating the array's items schema. This caused nested arrays to lose their `Array()` wrapper.

**Core fix in `updateSchemaAtPath`**: When the target property is an array and we're passing a non-array schema, preserve the `Array()` wrapper (we're updating items, not replacing the array). Only replace entirely when passing an array schema (type change).

### Other changes / improvements
**Unified multi-value config handling**: Added helper functions that handle the implicit root array wrapper for both RANDOM_PERMUTATION and BALANCED_ASSIGNMENT:
   - `isMultiValueConfig()` - type guard for configs with multiple values
   - `getItemSchemaIfArray()` - extracts item schema from array (DRY)
   - `updateSchemaForConfig()` - unwraps root array, updates, rewraps
   - `setValueForConfig()` - operates on item schema for value updates
   - `renamePropertyForConfig()` - operates on item schema for renames

3. **Consistent UI**: For multi-value configs at root, extract the item schema and show "Type: X" directly instead of "Array items: X". Uses `rootArrayUnwrapped` flag to prevent over-extraction for nested arrays.

### Limitations

Deeply nested arrays (3+ levels) may not update correctly because the path doesn't change when entering array items. A future fix could add `[]` notation to paths and handle it in `updateSchemaAtPath`.

### DIFF
--- a/frontend/src/components/experiment_builder/variable_editor.ts
+++ b/frontend/src/components/experiment_builder/variable_editor.ts
@@ -20,7 +20,7 @@ import {ExperimentEditor} from '../../services/experiment.editor';
 
 import {
   type BalancedAssignmentVariableConfig,
-  type RandomPermutationVariableConfig,
+  type MultiValueVariableConfigType,
   type StaticVariableConfig,
   type VariableConfig,
   BalanceAcross,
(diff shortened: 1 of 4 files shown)


jimbojw:
I don't have any notes for this PR. The code looks good to me, but I'm still onboarding. I did look to see if/how this could relate to #909. They don't _look_ related to me, but I don't fully grok the config variable code, so there may be an interaction I don't yet see.

PR #955, opened 2026-01-14
Remove legacy prompt types and improve schema validation

rasmi:
My original goal was to make the API options optional, but then I ran into SO MANY validation issues when updating the js/python schemas, and found some old types that were lingering in the codebase. this PR removes those types (and also now makes chat API options, optional).

**Made `ModelGenerationConfig` fields optional:**
- Fields like `maxTokens`, `temperature`, `topP` are now optional
- When unset, the API provider uses its own defaults instead of hardcoded values
- `functions/src/api/ai-sdk.api.ts` updated to only pass fields that are explicitly set

___
## Validation / Type changes

Removed the legacy `BaseAgentPromptConfig` prompt system (which used `promptContext: string`) in favor of the current `BasePromptConfig` system (which uses `prompt: PromptItem[]`). Also added TypeBox validation schemas to types that were missing it.

### Background

The codebase had two parallel prompt systems from an earlier migration. The legacy system was only used by the `testAgentConfig` endpoint and some dead code. The database already uses the current format, making the legacy TypeScript types obsolete (i.e. codebase has strictly been using the new schema for some time now).

### Changes

**Removed legacy types** (already replaced by current system in `structured_prompt.ts` from #561):
- `BaseAgentPromptConfig` → replaced by `BasePromptConfig`
  - `promptContext: string` → `prompt: PromptItem[]`
- `AgentChatPromptConfig` → replaced by `ChatPromptConfig`
- `AgentPromptSettings` → fields moved into `BasePromptConfig`:
  - `numRetries` → `BasePromptConfig.numRetries`
  - `includeStageHistory` → handled by `StageContextPromptItem` in prompt array
  - `includeStageInfo` → `BasePromptConfig.includeScaffoldingInPrompt`
- `AgentDataObject`, `AgentChatResponse` (dead code, never imported)
- `createAgentPromptSettings()` factory function

**Updated test endpoint** to use `prompt: PromptItem[]`:
- `functions/src/agent.endpoints.ts` - converts PromptItem array to string
- `frontend/src/components/experimenter/experimenter_data_editor.ts` - uses `createDefaultPromptFromText()`

**Fixed stale type cast** in `functions/src/mediator.utils.ts`:
- Changed `AgentChatPromptConfig` → `MediatorPromptConfig` (was just using `.id` field)

**New and updated validation schemas:**
- `utils/src/prompt.validation.ts` **(new)** - PromptItem and PromptConfig schemas
- `utils/src/providers.validation.ts` **(new)** - ModelGenerationConfig, provider options schemas
- `utils/src/structured_output.validation.ts` **(new)** - StructuredOutputConfig schemas
- `utils/src/agent.validation.ts` **(updated)** - Added `AgentModelSettingsData`, proper `PromptConfigData` references for templates, removed legacy `AgentDataObjectData`
- `utils/src/stages/stage.validation.ts` **(updated)** - Added `StageKindData` enum schema
- `utils/src/variables.validation.ts` **(updated)** - Added `$id` to JSON schema types (`JSONSchemaString`, `JSONSchemaObject`, etc.) and `ShuffleConfig`

### Note

`ProfileAgentConfig.promptContext` was **intentionally kept** - it serves a different purpose (per-participant persona context) than the removed `BaseAgentPromptConfig.promptContext` (stage prompt content).


utils/src/agent.ts
```diff
@@ -214,13 +214,13 @@ export function createModelGenerationConfig(
...
 ): ModelGenerationConfig {
   return {
-    // Universal settings
-    maxTokens: config.maxTokens ?? 8192,
-    stopSequences: config.stopSequences ?? [],
-    temperature: config.temperature ?? 1.0,
-    topP: config.topP ?? 1.0,
-    frequencyPenalty: config.frequencyPenalty ?? 0.0,
-    presencePenalty: config.presencePenalty ?? 0.0,
+    // Universal settings (optional - let API use defaults if not specified)
```
cjqian: Is it important to you to remove defaults here? We recently completed a study where we compared different models and knew they were all at temperature 1 b/c it wasn't instantiated; this would cause a little less certainty

rasmi: You can always still set the desired values for a given experiment, the way it works now, you can't NOT set the temperature and leave it up to the provider. WDYT?

PR #966, opened 2026-01-21
Small chat UI enhancements

rasmi:
- **Add option to prevent message cancellation in private chat** - Experimenters can now disable the cancel button to prevent participants from gaming minimum message counts by repeatedly cancelling requests
- **Improve typing indicator UX** - Replace "Waiting for a response..." text with a visual spinner around the agent's avatar icon
- **Fix Enter key behavior** - Send messages immediately on Enter (without inserting newline first); use Shift+Enter for multi-line
- **Fix MobX reactivity for private chat messages** - Sometimes chat messages required a page refresh (maybe due to mobx observability not being triggered when mutating existing array rather than replacing it)?

### DIFF
--- a/frontend/src/components/stages/group_chat_participant_view.scss
+++ b/frontend/src/components/stages/group_chat_participant_view.scss
@@ -88,10 +88,7 @@ stage-footer {
 
 .spinner {
   position: absolute;
-  top: -3px;
-  left: -3px;
-  right: -3px;
-  bottom: -3px;
+  inset: -3px;
   border: 2px solid transparent;
   border-top-color: var(--md-sys-color-primary);
   border-radius: 50%;


frontend/src/components/chat/chat_input.ts
```diff
@@ -48,10 +48,10 @@ export class ChatInputComponent extends MobxLitElement {
       this.isLoading = false;
     };
 
-    const handleKeyUp = async (e: KeyboardEvent) => {
-      if (e.key === 'Enter') {
+    const handleKeyDown = async (e: KeyboardEvent) => {
```
jimbojw: I think I understand your rationale here. By calling `e.preventDefault()` on `keydown`, you keep the browser from inserting a literal newline on the Enter-to-send.

The downside of `keydown` is that it calls `sendInput()` the instant the user presses down on the Enter button, if they weren't already pressing Shift.

My question would be whether trailing newlines are ever desirable in chat inputs? If not, then we could stick with `keyup` and trim trailing newlines (or all trailing whitespace), which would allow the user to press Shift and Enter in either order.

But if trailing newlines _are_ valid, then this current implementation is probably fine. The `keyup` way to handle it would have to somehow delete exactly one trailing newline if Shift was not pressed at `keyup` time.

It would be tempting to use both a `keydown` and a `keyup`, so that the `keydown` would prevent the newline, and the `keyup` would call `sendInput()`. But unfortunately that won't work because, `e.preventDefault()` not only cancels the newline insertion, it also prevents later synthetic events from firing. That is, in the split case, there wouldn't be a `keyup` event at all.

rasmi: I tried it, but it results in the same glitchy UI behavior where the chat bar expands and then contracts from the newline unfortunately!

frontend/src/components/stages/group_chat_participant_view.scss
```diff
@@ -71,3 +71,38 @@ stage-footer {
...
+  min-height: calc(common.$spacing-xxl * 1.5);
+}
+
+.spinner {
+  position: absolute;
+  top: -3px;
+  left: -3px;
+  right: -3px;
+  bottom: -3px;
+  border: 2px solid transparent;
```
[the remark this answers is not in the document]
rasmi: Desired behavior! 
Screenshot 2026-01-22 at 3 17 26 PM

PR #976, opened 2026-01-27
Update Python client to require specifying dev vs. prod.

rasmi:


jimbojw (approved):
I'm not a fan of having production info committed to the code base, but this was an existing condition, so your PR is an improvement.

PR #981, opened 2026-01-30
Update StockInfo and AssetAllocation to support resolving stock IDs 

rasmi:


frontend/src/components/stages/asset_allocation_participant_view.ts
```diff
@@ -58,13 +59,46 @@ export class AssetAllocationParticipantView extends MobxLitElement {
...
+  /** Stocks with ID-based lookup and failsafe parsing */
   private get stocks() {
     if (!this.stage) {
       return {stockA: null, stockB: null};
     }
+
+    const config = this.stage.stockConfig;
+
+    // Failsafe: ensure parsedData exists
+    const ensureParsedData = (stock: Stock | undefined): Stock | null => {
```
jimbojw: Nit: I believe this could be shortened to:

```ts
const ensureParsedData = (stock?: Stock): Stock | null => {
```

You don't have to change it, just noting that `| undefined` is what I believe the `?` means.

rasmi: Done!

utils/src/stages/stockinfo_stage.manager.ts
```diff
@@ -1,14 +1,43 @@
...
+      );
+      return {...updatedStage, visibleStockIds: resolvedIds};
+    }
+
+    return updatedStage;
+  }
+
   getStageDisplayForPrompt(
-    participants: ParticipantProfileExtended[],
+    _participants: ParticipantProfileExtended[],
```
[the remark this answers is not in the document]
rasmi: Done, ty!

PR #984, opened 2026-01-31
Update AssetAllocation design

rasmi:
(Depends on #981)

Update StockInfo and AssetAllocation to allow an InitialInvestment parameter, which is then used for growth charts in StockInfo and also for numeric values in AssetAllocation. Also allows overriding StageFooter text.

### DIFF
--- a/frontend/src/components/stages/asset_allocation_participant_view.scss
+++ b/frontend/src/components/stages/asset_allocation_participant_view.scss
@@ -1,4 +1,5 @@
 @use '../../sass/common';
+@use '../../sass/typescale';
 
 :host {
   @include common.flex-column;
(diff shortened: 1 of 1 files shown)


jimbojw:
This looks substantially similar to #981?

jimbojw (commented):
I like that we're expanding this to work with different currencies other than USD. My main recommendation would be to consolidate the currently repeated currency formatting logic into a single utility function, and then use [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) with `{ style: 'currency' }` therein to allow the browser to perform the formatting.

You may consider using `'USD'` as the currency specifier string rather than `'$'` in your data structures. This would remove the need to locale check the Indian currency symbol, and make it forward compatible with any other future currency strings.

frontend/src/components/stages/asset_allocation_participant_view.scss
jimbojw: Thank you for switching to `@include typescale` and `var(--md-...)`. 🙏

PR #996, opened 2026-02-05
Add frontend deployment steps to Google Cloud Build config

jimbojw:
Adds Deliberate Lab frontend to Google Cloud Build config (`cloudbuild.yaml`), including:

* Substitution values needed by the frontend
* Validation of substitution values
* Performing frontend unit tests
* Build frontend
* Deploy to Google Cloud App Engine

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -414,9 +414,9 @@ steps:
   waitFor: ['TestFunctions']
   dir: '.'
 
-# Deploy Rules Step
+# Deploy Firestore Rules Step
 - name: 'node:22'
-  id: DeployFirebaseRules
+  id: DeployFirebaseFirestoreRules
   env:
     - 'DEPLOYMENT_TYPES=" all rules "'
   script: |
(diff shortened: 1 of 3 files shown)


frontend/index.example.html
```diff
@@ -1,34 +1,42 @@
...
         // Create a script element for the config code
         var configScript = document.createElement('script');
 
-        // TODO: Replace 'your-google-analytics-id' with real ID
         configScript.textContent = `
           window.dataLayer = window.dataLayer || [];
           function gtag(){dataLayer.push(arguments);}
           gtag('js', new Date());
-          gtag('config', 'your-google-analytics-id');
+          gtag('config', '<%= analyticsId %>');
```
jimbojw: Same here. This is the only other substantive change. All other changes are formatting only.

frontend/webpack.config.mts
jimbojw: Here, we read the environment variable `MEASUREMENT_ID` and use it if supplied.

If not supplied, then there's no difference to just using `index.example.html` as is. This approach allows us to keep all the existing documentation as-is (it's all still correct, in that it works).

But, rather than hard coding the analytics ID in the `index.html`, this approach allows that secret to be injected in at build time, which the Google Cloud Build does.

cloudbuild.yaml
rasmi: (maybe add the comment here too that it doesn't include frontend on purpose -- though technicall the presubmit hooks and tests should have caught it by now!)

rasmi (approved):

cloudbuild.yaml
jimbojw: Nice catch! That line was leftover from the first phase of the work when Google Cloud Build only did backend services.

Since the Cloud build now also does frontend, I've separated out LintCheck into LintUtils, LintFunctions and LintFrontend.

frontend/webpack.config.mts
```diff
@@ -76,6 +76,9 @@ export default (
         favicon: './favicon.png',
         // Prepend an optional prefix path to the base URL of referenced assets in index.html
         base: process.env.URL_PREFIX ?? '/',
+        templateParameters: {
+          analyticsId: process.env.MEASUREMENT_ID || 'your-google-analytics-id',
+        },
       }),
       new webpack.DefinePlugin({
         'process.env.URL_PREFIX': process.env.URL_PREFIX ?? "'/'",
```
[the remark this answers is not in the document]
rasmi: It's not so serious! Maybe use `node -e` instead of `echo` when writing out those files to get the character escaping, in case those configs ever end up with weird characters (or maybe we want them to fail loudly in such cases).

PR #912, opened 2025-12-12
Improve API key testing.

mkbehr:
## Description
Improve the API key testing functionality:
- Error responses now properly appear as errors.
- Error responses display detailed error messages.
- OpenAI and Claude APIs now use GPT 5 Nano and Claude 4.5 Haiku for testing, instead of trying to test with Gemini.
  - This is hardcoded, so experimenters using custom endpoints can't specify which model to test with.

- [x] [Documentation](https://pair-code.github.io/deliberate-lab/) updated as needed (n/a)

---

## Verification
- [x] *Screenshots or videos* included (if applicable)
- [x] Clear reproduction steps provided to invoke the feature (if applicable) (n/a)
- [x] All tests pass (if applicable)

Tested manually with Gemini and OpenAI keys.

Screenshot 2025-12-12 at 4 18 10 PM

---

## Related issues
This PR fixes: #884 

---

PR #997, opened 2026-02-07
fix: Use TestAgentPromptConfig type

jimbojw:
Update `testAgentConfig()` method to use `TestAgentPromptConfig` type.

### DIFF
--- a/frontend/src/services/experiment.manager.ts
+++ b/frontend/src/services/experiment.manager.ts
@@ -1038,7 +1038,7 @@ export class ExperimentManager extends Service {
   /** Test given agent config. */
   async testAgentConfig(
     agentConfig: AgentPersonaConfig,
-    promptConfig: BaseAgentPromptConfig,
+    promptConfig: TestAgentPromptConfig,
   ): Promise<ModelResponse> {
     const creatorId = this.sp.authService.experimenterData?.email;
     if (!creatorId) {


rasmi:
#912 didn't properly merge / rebase the changes from the chat refactor (looks like this just happened yesterday): https://github.com/PAIR-code/deliberate-lab/commit/7823b270ab22d9b994af4bf5483a7e2534a249f1

@mkbehr -- there was a pretty substantial chat refactor, I think #912 can be simplified a bit!

jimbojw:
Thanks Rasmi, you're right, that was the cause of the regression.

I think I found the root cause that prevented GitHub Actions from detecting the issue. Our `ci.yaml` runs `npm test` for the frontend workspace, BUT it doesn't actually run `npm build`. It _does_ build both utils and functions, just not frontend.

So, if I understand correctly, as long as the frontend tests only touch buildable code, they can run and pass. Since the offending line was in a file outside the tested files' dependency trees, the build failure wasn't caught.

I'm writing up a GitHub Issue to track this CI enhancement/bug.

rasmi:
I wonder if we should adjust the CI commands so that test is always build+test (never just test alone)? Or just add a separate build step like the others have?

jimbojw:
@rasmi As I see it, for this issue (the `TestAgentPromptConfig` regression) we have three paths forward:

1. Merge this PR to fix the immediate build problem.
2. Rollback #912 to fix the immediate build problem.
3. Develop a new PR to address the simplifications you mentioned that should now be possible.

Thoughts?

rasmi (approved):

PR #1004, opened 2026-02-11
Add codelab for experiment creation using the Deliberate Lab API

aldenhallak:
## Description

- [x] [Documentation](https://pair-code.github.io/deliberate-lab/) updated as needed.

I gave my API key to gemini and let it run through the codelab and it successfully generated 
image

---

## Related issues

1005  -- added a default discussions [] field for the API.
---


rasmi:
Worth linking this here! https://pair-code.github.io/deliberate-lab/developers/api
And FYI types.py is meant to be auto-generated, never modified manually -- I can add a lint check!

PR #1008, opened 2026-02-12
Misc null fixes

rasmi:
* Add null fallback for chatStageConfig.discussions.
* Cohort deletion didn't mark participants as deleted before removing the cohort document, causing a race condition where participant triggers would crash trying to access the now-missing cohort's stageUnlockMap.

### DIFF
--- a/utils/src/stages/chat_stage.ts
+++ b/utils/src/stages/chat_stage.ts
@@ -161,9 +161,7 @@ export function createChatStagePublicData(
   config: ChatStageConfig,
 ): ChatStagePublicData {
   const id = config.id;
-  const discussions = config.discussions ?? [];
-  const currentDiscussionId =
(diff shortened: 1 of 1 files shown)


utils/src/stages/chat_stage.ts
```diff
@@ -161,8 +161,9 @@ export function createChatStagePublicData(
   config: ChatStageConfig,
 ): ChatStagePublicData {
   const id = config.id;
+  const discussions = config.discussions ?? [];
   const currentDiscussionId =
-    config.discussions.length === 0 ? null : config.discussions[0].id;
+    discussions.length === 0 ? null : discussions[0].id;
```
jimbojw: Nit: I'm pretty sure one could get away with the following one liner:

```ts
const currentDiscussionId = config.discussions?.[0]?.id ?? null;
```

PR #1015, opened 2026-02-18
Add Logs export to API

rasmi:
* Add Logs export to API
* Bump RAM for API endpoint
* Update API docs.


jimbojw:
The description mentions updating API docs. Does this mean inline documentation, like code comments, or should there be Markdown file updates that aren't yet present?

jimbojw (approved):

PR #1054, opened 2026-03-11
Show participant number for anonymous_animal, add informal name formatting

rasmi:
Closes #1048

Always include random numeric ID in anonymous animal profile names, making them look more like usernames (e.g., "Bear 1002") instead of the current format which only appends a sequential number on collision (e.g., "Bear", "Bear 2"). This aligns behavior with ANONYMOUS_PARTICIPANT appraoch (e.g. "Participant 1002") and uses the same random number already generated for the public ID and anonymous participant profiles. This is now default behavior for new experiments going forward.

Also: 
* Changes the quickstart to use anonymous animal profiles instead of open-form profiles.
* Adds a new `informalNameStyle` option which displays names as `bear123`.
* Nicer profile stage editor UI.

Screenshot 2026-03-11 at 3 53 30 PM

### DIFF
--- a/docs/assets/api/schemas.json
+++ b/docs/assets/api/schemas.json
@@ -1450,10 +1450,6 @@
               "type": "string"
             }
           ]
-        },
-        "showParticipantNumber": {
-          "default": true,
-          "type": "boolean"
         }
       }
     },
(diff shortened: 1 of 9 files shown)


mkbehr (changes requested):

jimbojw (approved):
Sorry for the delay. Looks good. Please confirm it still works given intervening PRs.

PR #1057, opened 2026-03-12
More clear AgentMessageResult handling to fix stuck AgentParticipants (#1011)

### LINKED ISSUE #1011: Agent participants can get stuck in private chat
I was testing agent participants today, and had one get stuck in a private chat stage. This left the experiment stuck until I manually hit the next stage button for the agent.

> Mediator 16:31
> Hello! I'm the Mediator. Since we're in this private chat, I'd love to get to know you a bit better. To start things off, what's your favorite flavor of ice cream?
> 
> AI Assistant (it/its) 16:31
> Hello Mediator! It's good to meet you. My favorite flavor of ice cream would have to be classic vanilla bean.
> 
> Mediator 16:31
> Vanilla bean is a classic choice! Thank you for sharing that with me.
> 
> AI Assistant (it/its) 16:31
> You're welcome! I appreciate you asking. I'm happy to share.

Checking the logs, the mediator here gave an empty response with `readyToEndChat: true`, at which point the agent didn't get any more queries.

Some ideas for options that could fix this:
1. End private chat stages when the mediator leaves or runs out of questions
2. Don't let the mediator choose to end a private chat stage
3. Wake agents up when someone leaves the chat
4. End any chat stage when there are no humans left and no pending agent responses
5. Wake agents up every minute or so if the chat has been idle
6. Auto-end chat stages if they're idle for too long
7. Default the private chat timer time limit to on

rasmi:
Fixes #1011

Agent participants get stuck indefinitely in private chat stages when all mediators stop responding. The trigger (`onPrivateChatMessageCreated`) is event-driven -- when no mediator writes a new message, no trigger fires, and the agent has no opportunity to advance.

A secondary issue: in private chats, `shouldRespond: false` unconditionally forced `readyToEndChat: true` on agent participants, conflating a per-turn skip with a conversation-ending signal.

### `functions/src/chat/chat.agent.ts`
- Add `AgentMessageResult` enum (`SENT`, `DECLINED`, `ERROR`) replacing boolean returns from `createAgentChatMessageFromPrompt`
- Decouple `shouldRespond` from `readyToEndChat` — respect the model's actual `readyToEndChat` value instead of forcing it
- Remove dead-code empty `if (!shouldRespond)` block from #846

### `functions/src/triggers/chat.triggers.ts`
- Collect mediator results and check if all mediators are done
- Advance agent participant when conversation is dead (all mediators declined)
- Guards: message type (prevent premature advancement on mediator self-decline), `currentStageId` (prevent double-advancement), `participant.agentConfig` (no effect on human participants)
- Skip "No mediators found" error for agent participants

## What this does NOT change
- Human participant behavior
- Group chat stages (return type change applies to shared code but group chat trigger discards the return value -- no change to existing group chat behavior)   
- Mediator behavior (return type change only)
- Frontend spinner handling

### DIFF
--- a/functions/src/triggers/chat.triggers.test.ts
+++ b/functions/src/triggers/chat.triggers.test.ts
@@ -1,0 +1,541 @@
+import firebaseFunctionsTest from 'firebase-functions-test';
+import {StageKind, UserType} from '@deliberation-lab/utils';
+
+// All mocks are built inside factories to avoid Jest hoisting issues.
+// Access them via __mocks__ re-exports after importing.
+
+jest.mock('../app', () => {
+  const getMock = jest.fn();
+  const setMock = jest.fn().mockResolvedValue(undefined);
+  const transactionSetMock = jest.fn();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const chain: Record<string, any> = {
+    get: getMock,
+    set: setMock,
+    collection: jest.fn(),
+    doc: jest.fn(),
+    runTransaction: jest.fn(async (fn: (tx: {set: jest.Mock}) => unknown) =>
+      fn({set: transactionSetMock}),
+    ),
+  };
+  chain.collection.mockReturnValue(chain);
+  chain.doc.mockReturnValue(chain);
+  return {
+    __esModule: true,
+    app: {
+      firestore: jest.fn().mockReturnValue(chain),
+    },
+    __mocks__: {
+      getMock,
+      setMock,
+      transactionSetMock,
+    },
+  };
+});
+
+jest.mock('../chat/chat.agent', () => {
+  const createAgentChatMessageFromPromptMock = jest.fn();
+  const actual = jest.requireActual('../chat/chat.agent');
+  return {
+    __esModule: true,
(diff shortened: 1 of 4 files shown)


functions/src/triggers/chat.triggers.test.ts
```diff
@@ -0,0 +1,541 @@
...
+        .mockResolvedValueOnce(AgentMessageResult.SENT);
+
+      await wrapped({params: EVENT_PARAMS});
+
+      expect(
+        participantUtilsMocks.updateParticipantNextStageMock,
+      ).not.toHaveBeenCalled();
+    });
+
+    it('does NOT advance when all mediators return error', async () => {
```
mkbehr: Is this the correct behavior? If an agent participant is chatting with a mediator and the mediator gets an error responding, won't we still get stuck? Or if that actually does trigger the agent participant to respond, then can we verify that in this test?

mkbehr (changes requested):
Thanks for the fix!

jimbojw (approved):
Sorry for the delay on this. Please make sure it still works as intended given intervening PRs.

PR #1079, opened 2026-04-10
Check for empty variable config when creating experiments

### LINKED ISSUE #1074: createCohort crashes when a RandomPermutation variable has an empty values array
## Description

Creating a cohort fails with a 500 Internal Server Error when the experiment contains a `RandomPermutationVariableConfig` with an empty `values` array.

## Root Cause

In [generateRandomPermutationValue()](cci:1://file:///usr/local/google/home/jimbo/checkout/deliberate-lab/utils/src/variables.utils.ts:212:0-262:1) ([utils/src/variables.utils.ts](cci:7://file:///usr/local/google/home/jimbo/checkout/deliberate-lab/utils/src/variables.utils.ts:0:0-0:0)), `numToSelect` is clamped to the range `[1, maxAvailable]`:

    const maxAvailable = config.values.length;  // 0
    const requestedNum = config.numToSelect ?? maxAvailable;  // 0
    const numToSelect = Math.max(1, Math.min(requestedNum, maxAvailable));  // Math.max(1, 0) → 1

When `values` is empty, `maxAvailable` is 0, so `Math.min(requestedNum, 0)` yields 0, then `Math.max(1, 0)` clamps it up to 1. The function then calls [choices(config.values, 1, seedValue)](cci:1://file:///usr/local/google/home/jimbo/checkout/deliberate-lab/utils/src/utils/random.utils.ts:106:0-131:2), which throws:

    Error: Cannot choose 1 distinct values from an array of length 0

This unhandled error propagates up through `createCohortInternal()` → `runTransaction()` and results in a generic 500 response to the client with no useful error message.

## Effect

- The [createCohort](cci:1://file:///usr/local/google/home/jimbo/checkout/deliberate-lab/frontend/src/services/experiment.manager.ts:794:2-820:3) cloud function returns a 500 Internal Server Error
- The client receives no indication of what went wrong—the browser console only shows `POST …/createCohort 500 (Internal Server Error)`
- The actual error is only visible in Cloud Functions logs (`firebase functions:log --only createCohort`)

### Option A: Handle the empty-values edge case gracefully

If `values` is empty and `numToSelect` is 0 (or unset), return an empty array instead of attempting to select:

    if (maxAvailable === 0) {
      if (requestedNum === 0 || requestedNum === undefined) {
        return [];
      }
      throw new HttpsError(
        'invalid-argument',
        `Variable "${config.definition.name}": cannot select ${requestedNum} values from an empty array`
      );
    }

### Option B: Add validation to reject empty values at save time

Add a check in the variable editor or the experiment write endpoint to prevent saving a `RandomPermutationVariableConfig` with an empty `values` array. The validation schema in [utils/src/variables.validation.ts](cci:7://file:///usr/local/google/home/jimbo/checkout/deliberate-lab/utils/src/variables.validation.ts:0:0-0:0) currently allows it since `values` has no `minItems` constraint.

### Option C: Both

Handle the edge case at generation time (Option A) for robustness, and also add validation at save time (Option B) so users get early feedback.

## Workaround

The user can edit the experiment in the Experiment Builder, find the variable (named "variable") in the Variables section, and either delete it or add at least one value to its values array.

mkbehr:
## Description

Add validation that checks for empty variables when creating an experiment. This applies to both the API and the web interface. Previously, an experiment could be saved with a bad variable config, but it would fail when trying to create a cohort.

---

## Verification

ArnphiaRS3w8of9

---

## Related issues
This PR fixes: #1074 

---

### DIFF
--- a/functions/src/experiment.utils.ts
+++ b/functions/src/experiment.utils.ts
@@ -328,13 +328,6 @@ export interface UpdateExperimentOptions {
  */
 export interface UpdateExperimentResult {
   success: boolean;
-  error?:
-    | 'not-found'
-    | 'not-owner'
-    | 'cohort-definitions-locked'
-    | 'invalid-stages'
-    | 'invalid-variables'
-    | 'duplicate-cohort-aliases';
   httpErrorCode?: number;
   errorMessage?: string;
 }
@@ -365,7 +358,6 @@ export async function updateExperimentFromTemplate(
   if (!stageValidation.valid) {
     return {
       success: false,
-      error: 'invalid-stages',
       httpErrorCode: 400,
       errorMessage: `Invalid stage configuration: ${stageValidation.error}`,
     };
@@ -379,7 +371,6 @@ export async function updateExperimentFromTemplate(
     if (!variableValidation.valid) {
       return {
         success: false,
-        error: 'invalid-variables',
         httpErrorCode: 400,
         errorMessage: `Invalid variable configuration: ${variableValidation.error}`,
       };
@@ -393,7 +384,6 @@ export async function updateExperimentFromTemplate(
       if (seenAliases.has(def.alias)) {
         return {
           success: false,
-          error: 'duplicate-cohort-aliases',
           httpErrorCode: 400,
           errorMessage: `Duplicate cohort alias found: "${def.alias}"`,
         };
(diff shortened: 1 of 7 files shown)


functions/src/experiment.utils.ts
aldenhallak: is this necessary / used anywhere?

aldenhallak: do we use error or just errorcode and errormessage? Should we get rid of Error?

aldenhallak (approved):

functions/src/experiment.utils.ts
mkbehr: Removed it.

PR #1088, opened 2026-04-21
Fix per-participant survey to map over participants instead of questions first

Aa-g-i:
Feedback from participants / playtesters said the current format for representing per-participant survey questions (each question, for all participants, rather then grouped by-participant) is unintuitive and difficult to follow with more then two per-participant survey questions. This PR represents a potential solution to that, grouping by participant (each question for that participant, then the next participant).


utils/src/variables.utils.ts
```diff
@@ -421,7 +421,7 @@ export function validateParsedVariableValue(
   const valid = ajv.validate(schema, value);
   if (!valid && ajv.errors) {
     const errorMsg = ajv.errors
-      .map((e) => `${e.instancePath || '/'}: ${e.message}`)
+      .map((e: any) => `${e.instancePath || e.dataPath || '/'}: ${e.message}`)
```
jimbojw: Is the `any` because of the `dataPath`? If so, I don't see `dataPath` anywhere else in the codebase?

frontend/src/components/stages/survey_per_participant_view.ts
```diff
@@ -325,7 +326,6 @@ export class SurveyView extends MobxLitElement {
           .minLength=${minCount ?? nothing}
           .maxLength=${maxCount ?? nothing}
           .error=${isTooShort}
-          .errorText=${errorText}
```
jimbojw, line 328: Was removing this intentional?

jimbojw: @Aa-g-i

jimbojw (commented):
Can you confirm that removing the `errorText` was intentional? Under what conditions would there previously have been an error? Are those conditions no longer appliciable?

Aa-g-i:
Reverted the error handling. It was leftover from something else. 
That something else is worth discussing: it seems as though either BOTH the local / deployed backend or just the local backend do not handle nulls well. Just fyi

jimbojw:
OK thanks. If you find a case where a null is not being handled well, I recommend making a GitHub issue for it. You don't have to come up with the text yourself, your AI agent can do that. Just ask it to give you the Markdown of a GitHub issue and it'll render a draft you can copy-paste.

jimbojw (approved):

PR #1095, opened 2026-04-24
Fix minimum-time in discussions so it actually blocks the "Next Stage" button until the minimum time has been passed. 

Aa-g-i:
Verified via local run.

### DIFF
--- a/functions/src/participant.endpoints.ts
+++ b/functions/src/participant.endpoints.ts
@@ -14,7 +14,6 @@ import {
   setProfile,
   VariableScope,
   getTimeElapsed,
-  ChatStageConfig,
   ChatStagePublicData,
 } from '@deliberation-lab/utils';
 import {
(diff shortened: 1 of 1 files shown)


functions/src/participant.endpoints.ts
```diff
@@ -366,6 +374,48 @@ export const updateParticipantToNextStage = onCall(
...
+            const messages = await getFirestorePrivateChatMessages(
+              data.experimentId,
+              participant.privateId,
+              currentStageId,
+            );
+            startTimestamp = messages.length > 0 ? messages[0].timestamp : null;
+          }
+
+          if (
+            !startTimestamp ||
```
jimbojw, line 415: As I read this, `startTimestamp` will be undefined any time `stage.kind` is not `CHAT` or `PRIVATE_CHAT`. Under the current logic, this will always trigger a `failed-precondition` error.

Are there other kinds of stages we should be accounting for?

functions/src/participant.endpoints.ts
```diff
@@ -366,6 +374,48 @@ export const updateParticipantToNextStage = onCall(
...
       ).data() as ParticipantProfileExtended;
 
+      // Check minimum time enforcement for chat stages
+      const currentStageId = participant.currentStageId;
+      const stage = await getFirestoreStage(data.experimentId, currentStageId);
+      if (
+        stage &&
+        (stage.kind === StageKind.CHAT || stage.kind === StageKind.PRIVATE_CHAT)
+      ) {
+        const minTime = (stage as ChatStageConfig).timeMinimumInMinutes;
```
jimbojw: Nit: I like that `timeMinimumInMinutes` has the unit baked into the name. However, `minTime` does not.

functions/src/participant.endpoints.ts
```diff
@@ -366,6 +374,48 @@ export const updateParticipantToNextStage = onCall(
...
+              (publicStageData as ChatStagePublicData)
+                ?.discussionStartTimestamp ?? null;
+          } else if (stage.kind === StageKind.PRIVATE_CHAT) {
+            const messages = await getFirestorePrivateChatMessages(
+              data.experimentId,
+              participant.privateId,
+              currentStageId,
+            );
+            startTimestamp = messages.length > 0 ? messages[0].timestamp : null;
+          }
```
jimbojw, line 412: Should there be an `else` fallthough here for other values of `stage.kind`?

Aa-g-i: Added a fallthrough here~

functions/src/participant.endpoints.ts
Aa-g-i: Fixed, good catch

functions/src/participant.endpoints.ts
Aa-g-i: Not currently; it makes sense to have minimum time for discussions, haven't heard anyone else request it. However, we added a fallback / fallthrough default logic block here...

jimbojw (approved):
Looks good. I'd be mildly interested in if there are tests that need updating, but not critical.

PR #1102, opened 2026-05-01
Range surveys: require integers and improve validation

### LINKED ISSUE #1098: Add additional validation on creating an experiment with bad multiple choice values
Currently, we allow creation of scale questions  with lower value greater than upper value

mkbehr:
## Description

- Potentially breaking change: require numbers in range-type surveys to be integers
  - From manual testing, this already wasn't supported (at least for the web editor), so make that explicit.
- Check range-type survey stages for various requirements:
  - min < max
  - step size > 0
  - step size | max - min
  - all numbers are integers
- Require survey stages to contain at least one question
- Validate experiments in edit mode as well as creation mode
- Change response code for bad experiment config from 400 to 422 (more specific)

---

## Verification

Screenshot 2026-05-01 at 3 59 48 PM

---

## Related issues
This PR fixes: #1098

---


functions/src/dl_api/experiments.dl_api.ts
```diff
@@ -214,10 +216,12 @@ export async function createExperiment(
...
-    throw createHttpError(
-      400,
-      error instanceof Error ? error.message : String(error),
-    );
+    const msg = error instanceof Error ? error.message : String(error);
+    const isValidationError =
+      msg.includes('Invalid stage configuration') ||
+      msg.includes('Invalid variable configuration') ||
+      msg.includes('Duplicate cohort alias found');
+    throw createHttpError(isValidationError ? 422 : 500, msg);
```
jimbojw, line 224: Thanks for upgrading from `400` to more specific HTTP error codes. 🙏

frontend/src/services/experiment.editor.ts
```diff
@@ -118,17 +120,70 @@ export class ExperimentEditor extends Service {
...
+              !Number.isInteger(question.lowerValue) ||
+              !Number.isInteger(question.upperValue) ||
+              (question.stepSize !== undefined &&
+                !Number.isInteger(question.stepSize))
+            ) {
+              errors.push(
+                `${stage.name} question ${index + 1} ("${question.questionTitle}"): values must be integers`,
+              );
+            }
+            if (question.lowerValue >= question.upperValue) {
```
jimbojw: My main nit is that here and elsewhere, you assert a condition that would make the comparison potentially unspecified.

For example, earlier, you check `!Number.isInteger(question.lowerValue)` and if so, you append to the `errors` array. This is good. But in that case, when we get to this line, there are several possibilities, only some of which make sense. E.g., it could be that `lowerValue` is, say a float like `3.5`, or even a string like `"456"`. Those are legal values for this JavaScript `>=` comparison against `upperValue`, but they would have been flagged already as  being illegal (non-integer).

Anyway, if the code works and the tests pass, it's good enough for me. Just flagging that it's strange to me to see later comparisons using potentially bad values that were checked earlier.

frontend/src/services/experiment.editor.ts
```diff
@@ -118,17 +120,70 @@ export class ExperimentEditor extends Service {
...
+              );
+            }
+            const range = question.upperValue - question.lowerValue;
+            const step = question.stepSize ?? 1;
+            if (step <= 0) {
+              errors.push(
+                `${stage.name} question ${index + 1} ("${question.questionTitle}"): step size must be greater than 0`,
+              );
+            }
+            if (range % step !== 0) {
```
jimbojw: Same here. Above we check that `step` is positive. But here, we're outside that check. So we could be comparing a negative step number's modulus against the range. It's strange to me to allow these potentially weird comparisons, given that earlier validation checks aim to ensure correct types and values.

frontend/src/services/experiment.editor.ts
```diff
@@ -118,17 +120,70 @@ export class ExperimentEditor extends Service {
...
+              errors.push(
+                `${stage.name} question ${index + 1} ("${question.questionTitle}"): step size must divide max-min (${range}) exactly`,
+              );
+            }
+          }
+          if (question.kind === 'mc') {
```
jimbojw: Nit: Is there a static variable for the string `'mc'` we could use instead of the literal, such as an Enum value?

jimbojw (approved):

frontend/src/services/experiment.editor.ts
mkbehr: Good call. I lifted this check out into a separate function so we could early-return on failures.

frontend/src/services/experiment.editor.ts
mkbehr: Changed to `SurveyQuestionKind.MULTIPLE_CHOICE`.

PR #1115, opened 2026-05-12
Fix REST API validation failures for omitted fields

rasmi:
Commit f27f7fce introduced strict TypeBox `Value.Check` validation for REST API endpoints. This caused Python client payloads (which omit fields rather than explicitly sending `null` due to Pydantic's `exclude_none=True`) to be rejected with `Expected union value` 400 errors for any fields defined strictly as `Type.Union([Type.Null(), ...])`. 

This commit fixes the mismatch by wrapping all nullable TypeBox schemas in `Type.Optional()`, allowing clients to perform partial updates or omit empty fields cleanly while maintaining strict runtime typing.

### DIFF
--- a/docs/assets/api/schemas.json
+++ b/docs/assets/api/schemas.json
@@ -5,12 +5,7 @@
     "CohortParticipantConfig": {
       "title": "CohortParticipantConfig",
       "type": "object",
-      "required": [
-        "minParticipantsPerCohort",
-        "maxParticipantsPerCohort",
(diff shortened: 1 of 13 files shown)


cjqian (approved):

jimbojw:
I created #1118 to track the upstream problem. Is there an easy way to test for this at PR time? I believe we already run Firebase emulators for testing the functions. We should be able to also exercise the Python CLI at this time.

PR #1137, opened 2026-06-01
ci: Record previous configurations during Cloud Build deployment (#995)

### LINKED ISSUE #995: CI: Record previous rules on deploy
Currently, the [`cloudbuild.yaml`](https://github.com/PAIR-code/deliberate-lab/blob/main/cloudbuild.yaml), contains steps for deploying Deliberate Lab's various backend Firebase services including functions, indexes and rules. For indexes specifically, the deployment step makes note of the existing rules, then compares deployed rules to see if anything has changed.

However, the `DeployFirebaseRules` step does not perform these logging and checking substeps. As a result, the deployment _may overwrite_ deployed rules that were not specified in code. Firebase offers no historical view for changed rules, so this destructive operation is currently unmitigated.

To mitigate this potential data loss, we should enhance the `cloudbuild.yaml` to record and diff deployed rules just like it does for indexes.

jimbojw:
Currently, there are four different Firebase deployment steps. Of them, only updating indexes records the before and after state.

This change upgrades the deployment process so that all four Firebase deployment steps record the before and after state of deployed configs. That way, if there is ever a Firebase-based deployment issue, the Cloud Build logs will show what the configurations were just prior to the breaking change.

Fixes #995

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -102,6 +102,16 @@ steps:
         attempt=$(( attempt + 1 ))
       done
     }
+
+    # Fetch a GCP access token from the instance metadata server.
+    # Outputs the raw token string to stdout.
+    # Usage: local token; token=$(fetch_gcp_token)
+    fetch_gcp_token() {
+      local token_json
+      token_json=$(retry curl -sSf -H "Metadata-Flavor: Google" \
+        "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token")
+      node -p 'JSON.parse(require("fs").readFileSync(0,"utf-8")).access_token' <<< "$token_json"
+    }
     SHARED_SCRIPT
     chmod +x shared.sh
     echo "shared.sh written successfully."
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -473,7 +533,46 @@ steps:
     #!/usr/bin/env bash
     set -euo pipefail
     trap 'echo "❌ STEP FAILED: Command \"$BASH_COMMAND\" exited with code $?"' ERR
+
+    # Retry helper for transient network failures (read-only commands only).
+    retry() {
```
nthain: Is this retry copy-pasted a bunch of times? Is there an alternative?

nthain (approved):

cloudbuild.yaml
jimbojw: It is copied yeah. I'm not a fan of the copying either. I have an idea about how to streamline it.

jimbojw:
@nthain please take another look. I refactored the shared logic into a shell script that gets written as the first step of the build and then re-imported in each step that needs it.

This way, the `cloudbuild.yaml` continues to be the single source of truth for the build steps, but the repeated logic is collapsed into a single spot.

nthain (approved):

PR #1146, opened 2026-06-09
feat(participant): expose variables via SYSTEM_VARIABLE_NAMESPACE

jimbojw:
The effect of this would be a "virtual stage" under which participant-specific variables could be mapped to, which would allow us to do things like cohort Transfer based on the specific policy a participant was assigned in the policy_mediation experiment (as opposed to now, where we need to compartmentalize each policy into its own experiment).

### DIFF
--- a/functions/src/participant.utils.ts
+++ b/functions/src/participant.utils.ts
@@ -28,6 +28,7 @@ import {
   RankingStagePublicData,
   RoleStagePublicData,
   StageConfig,
+  SYSTEM_VARIABLE_NAMESPACE,
   STAGE_KIND_REQUIRES_TRANSFER_MIGRATION,
   StageKind,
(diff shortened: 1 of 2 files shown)


functions/src/participant.utils.ts
```diff
@@ -1263,6 +1264,29 @@ function buildTargetValuesForParticipant(
     }
   }
 
+  if (participant.variableMap) {
+    for (const [varName, varValue] of Object.entries(participant.variableMap)) {
+      try {
+        const parsed = JSON.parse(varValue);
+        if (typeof parsed === 'object' && parsed !== null) {
```
aldenhallak, line 1271: I might be missing some context, but are we intentionally trying to filter out JSON primitives here? if parsed is a primitive this will evaluate to false

PR #1160, opened 2026-06-14
feat: allow multi-turn private reasoning for agents

jacyanthis:
# Description
Adds multi-turn private reasoning for AI agents. When the experimenter includes `{{_reasoning}}` in an agent's prompt, the system adds an output field `_reasoning` to that agent's API queries. The agent is prompted to use the field for private reasoning. This allows the agent to make multi-turn plans without including them in their message.

For example, an agent could ask a participant, "Are apples your favorite food?" with the 'intention' of accounting for that information in a follow-up offer of apple pie for a lower price (if not favorite) or higher price (if favorite).

This feature does not affect functionality if the magic ``{{_reasoning}}`` placeholder is not used.

These reasoning traces are available in the database for future use. An explicit `_reasoning` field is used because not all models make their implicit reasoning traces available.

# Example
Example agent mediator prompt written by experimenter:

```
Your task is to mediate the conversation. {{_reasoning}}
```

Example prompt received by the model in the third round of a chat:

```
Your task is to mediate the conversation. Each line shows your reasoning for your messages sent in this conversation, in order:

#1
I will send the user a polite greeting. Then, before offering them the apple pie, I will ask if apples are their favorite food, so I can change my price accordingly.

#2
I will ask the user if apples are their favorite food, so I can change my price accordingly.
```

### DIFF
--- a/frontend/src/components/chat/chat_message.ts
+++ b/frontend/src/components/chat/chat_message.ts
@@ -68,7 +68,7 @@ export class ChatMessageComponent extends MobxLitElement {
   };
 
   override render() {
-    if (!this.chat || this.chat.isReasoningOnly) {
+    if (!this.chat || this.chat.isScratchpadOnly) {
       return nothing;
     }
 
--- a/frontend/src/components/chat/chat_message.ts
+++ b/frontend/src/components/chat/chat_message.ts
@@ -68,7 +68,7 @@ export class ChatMessageComponent extends MobxLitElement {
   };
 
   override render() {
-    if (!this.chat) {
+    if (!this.chat || this.chat.isReasoningOnly) {
       return nothing;
     }
 
--- a/frontend/src/services/cohort.service.ts
+++ b/frontend/src/services/cohort.service.ts
@@ -371,6 +371,7 @@ export class CohortService extends Service {
                 explanation: '', // backwards compatibility pre version 16
                 ...doc.data(),
               } as ChatMessage;
+              if (message.isReasoningOnly) return;
               if (!message.discussionId) {
                 this.chatMap[stageId].push(message);
               } else {
--- a/frontend/src/services/cohort.service.ts
+++ b/frontend/src/services/cohort.service.ts
@@ -371,7 +371,7 @@ export class CohortService extends Service {
                 explanation: '', // backwards compatibility pre version 16
                 ...doc.data(),
               } as ChatMessage;
-              if (message.isReasoningOnly) return;
+              if (message.isScratchpadOnly) return;
               if (!message.discussionId) {
                 this.chatMap[stageId].push(message);
               } else {
--- a/frontend/src/services/experiment.editor.ts
+++ b/frontend/src/services/experiment.editor.ts
@@ -200,10 +200,10 @@ export class ExperimentEditor extends Service {
     }
 
     for (const config of this.experiment.variableConfigs ?? []) {
-      // Block save and inform experimenter to avoid reasoning history name collision.
-      if (config.definition.name === '_scratchpad') {
+      // Names starting with "_" are reserved for system variables.
+      if (config.definition.name.startsWith('_')) {
         errors.push(
-          'The variable name "_scratchpad" is reserved for tracking the system\'s reasoning history. Please rename it.',
+          'Variables that begin with an underscore are not allowed unless they are established system variables.',
         );
       }
       if (requiresValues(config.type)) {
--- a/frontend/src/services/experiment.editor.ts
+++ b/frontend/src/services/experiment.editor.ts
@@ -17,6 +17,7 @@ import {
   VariableConfig,
   MultiValueVariableConfigType,
   requiresValues,
+  INTERNAL_VARIABLES,
   STAGE_MANAGER,
   checkApiKeyExists,
   SurveyQuestion,
@@ -200,8 +201,12 @@ export class ExperimentEditor extends Service {
     }
 
     for (const config of this.experiment.variableConfigs ?? []) {
-      // Names starting with "_" are reserved for internal variables.
-      if (config.definition.name.startsWith('_')) {
+      // Names starting with "_" are reserved for internal variables; only
+      // the ones in the registry may be used, and never defined here.
+      if (
+        config.definition.name.startsWith('_') &&
+        !INTERNAL_VARIABLES.has(config.definition.name)
+      ) {
         errors.push(
           'Variables that begin with an underscore are not allowed unless they are established internal variables.',
         );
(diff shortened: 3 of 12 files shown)


jacyanthis:
Possible review: @jimbojw 
CC: @kenburke @cjqian

jimbojw:
@jacyanthis Please resolve conflicts and re-upload.

jacyanthis:
I updated the branch with main, but it will likely become outdated again soon.

Manual testing per template:

Set up an experiment with (i) a chat stage and (ii) an agent mediator whose prompt includes `{{_reasoning}}`. The mediator should send at least one message and then be prompted to send another. The API prompt should show the reasoning from its previous messages. You could also test by eliciting a message that refers to its own reasoning (e.g., `Tell me exactly what you reasoned about before sending your previous message`), but that may be challenging.

jimbojw (changes requested):
## Naming & design feedback on the `_reasoning` field

Thanks for this — the private-notes-that-feed-back-into-the-prompt mechanism is an interesting capability.

I have some feedback on **naming and the surrounding architecture** that I think will make it clearer to future readers and easier to extend. Three recommendations, each with rationale.

---

### 1. Rename `_reasoning` → `_scratchpad`

**Recommendation:** Rename the feature (field, template variable, and structured-output field) from `_reasoning` to `_scratchpad`.

**Why:** This codebase already has a distinct **`reasoning`** concept — the model's native chain-of-thought / thinking tokens (`reasoningLevel`, `reasoningBudget`, `includeReasoning`, and the `reasoning` field on `ChatMessage`). The currently proposed `_reasoning` field means something quite different: an agent-authored private note that persists and is re-injected into later prompts. Right now the only thing distinguishing the two concepts is a leading underscore, which is very easy to misread — `msg.reasoning` vs `msg._reasoning` give no hint that they're fundamentally different things.

`scratchpad` is the industry-standard term for exactly this pattern (a private working area the agent writes to itself), and it has **no existing meaning in this codebase**, so it disambiguates cleanly with zero conceptual overlap.

Main touch-points for the rename:

- `utils/src/chat_message.ts` — the `_reasoning` field on `ChatMessage` (+ the `create*ChatMessage` factories). Also rename `isReasoningOnly` → `isScratchpadOnly`, since that flag describes this feature (not native reasoning).
- `utils/src/structured_output.ts` — the `_reasoning` field on `ChatMediatorStructuredFields`, `injectReasoningField`, and the model-facing field description string ("Your private reasoning for this turn…") should adopt scratchpad wording.
- `functions/src/structured_prompt.utils.ts` and `functions/src/chat/chat.agent.ts` — the `_reasoning` detection, value-map key, and history assembly.
- `frontend/src/services/experiment.editor.ts` — the reserved-name guard/message.

Since this changes exported `utils` types, don't forget `npm run update-schemas` so the schema-sync check passes.

---

### 2. Make the `_` prefix always mean "system variable" (and back it with a registry)

**Recommendation:** Reserve the entire leading-underscore namespace for system-defined variables, rather than special-casing the single literal `_reasoning`. Concretely:

- Treat any variable name beginning with `_` as **system-owned**.
- Block experimenter-defined variables whose names begin with `_` (generalize the current exact-match guard in `experiment.editor.ts` to a prefix check).
- Define the known system variables in a **single registry / set** (with one entry, `_scratchpad`, for now) and have the template logic consult it.

**Why:** Today the special handling for `_reasoning` is spread across several one-off string comparisons (e.g. `=== '_reasoning'` / `.includes('{{_reasoning}}')`) in `variables.template.ts`, `structured_prompt.utils.ts`, `structured_output.ts`, and `experiment.editor.ts`. That works for one variable, but every future system variable would mean adding another set of scattered checks. Reserving the `_` prefix and routing through a registry:

- **Affords future system variables** — adding a new one becomes a single registry entry instead of edits across four files.
- **Simplifies the current logic gates** — the various `_reasoning` checks collapse into "is this name in the system-variable registry?" / "does it start with `_`?".
- **Prevents collisions by construction** — experimenters can't define a variable that shadows (or is shadowed by) a system one.

---

### 3. Give undefined `{{_...}}` references a "system variable" hint

**Recommendation:** When a template references an unknown leading-underscore variable (e.g. a typo like `{{_scratchpat}}`), surface an error that says it's an **undefined *system* variable**, ideally listing the valid ones.

**Why:** The template validator already reports unresolved variables as `undefined`, so the basic error largely comes for free once `_scratchpad` is the only whitelisted system name — anything else `_`-prefixed falls through to the existing check. The improvement is just the *message*: because these are reserved, system-owned names, a generic "undefined variable" is less helpful than "`_scratchpat` is not a known system variable". This is especially valuable since a mistyped system variable would otherwise render as empty text with no other signal.

---

Happy to discuss any of these.

jimbojw:
## Follow-up: call these "internal variables", not "system variables"

Quick correction to my earlier comment. I recommended reserving the leading-`_` prefix for what I called "**system variables**." Please treat that naming as **superseded** — the concept should be called an "**internal variable**" instead. The mechanism (reserve the `_` prefix, back it with a registry, resolve/validate `{{_...}}` against it) is unchanged; only the *term* changes.

### Why the rename

"System variable" collides with **two existing concepts** already in this codebase, and reusing it would recreate exactly the kind of naming ambiguity the original feedback was trying to avoid:

1. **`SYSTEM_VARIABLE_NAMESPACE`** (`utils/src/utils/condition.ts`, value `'__system_variables__'`) — a sentinel pseudo-stage-ID used by the **condition system** to namespace participant `variableMap` values as comparison targets (see `functions/src/participant.utils.ts`). This is a different subsystem entirely (conditional logic, not template resolution).

2. **"generated variables"** (`functions/src/variables.utils.ts`) — already refers to *experimenter-defined* variables whose values are computed per participant (e.g. balanced assignment, shuffle). These are still authored by the experimenter; only their values are generated.

Both are close enough to cause confusion, so a distinct term is safer.

### Recommended vocabulary

- **internal variable** — the concept: a `_`-prefixed template token that the platform provides and the experimenter can reference but cannot define. ("Internal" also reinforces the leading-underscore convention, which already connotes internal/private.)
- **reserved** — the word for the *rule* and the validation message, e.g. *"names starting with `_` are reserved for internal variables."*
- **`_scratchpad`** — the first internal variable.

Suggested identifier shapes to keep it consistent: `INTERNAL_VARIABLES` (the registry/set), `InternalVariableName` (if a type is useful), and comments/docs referring to "internal variables."

No other part of the earlier recommendation changes — this is purely a terminology fix to avoid overloading "system"/"generated."

jacyanthis:
All four changes implemented, thanks!

Notes:
- I only blocked non-registered underscore-prefixed variables, rather than a more general block (since these have to be used).
- Putting `{{_scratchpat}}` in a prompt will not produce an error in the blocking-save sense, but it will show a warning that this is not a recognized internal variable (so it'll show up empty). The warning also states the underscore reservation is for internal variables because an experimenter might not know that.
- I also modified PR #1171 because that had its own registry system.

PR #1161, opened 2026-06-14
feat: allow conditional survey items to reference participant variables

jacyanthis:
# Description

Allows experimenters to make survey items conditionally appear not just based on responses to prior survey items (existing functionality) but also participant variables. For example, a survey question can only appear if a participant is in a certain treatment group. Variables appear under a protected "Variables" Stage ID.

# Example

image

### DIFF
--- a/frontend/src/components/experiment_builder/structured_prompt_editor.ts
+++ b/frontend/src/components/experiment_builder/structured_prompt_editor.ts
@@ -50,7 +50,11 @@ export class EditorComponent extends MobxLitElement {
     return getConditionTargetsFromStages(
       this.experimentEditor.stages,
       this.stageId,
-      {includeCurrentStage: true},
+      {
+        includeCurrentStage: true,
+        variableConfigs:
+          this.experimentEditor.experiment?.variableConfigs ?? [],
+      },
     );
   }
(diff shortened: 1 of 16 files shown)


jacyanthis:
Possible review: @jimbojw 
CC: @kenburke @cjqian

jimbojw:
@jacyanthis Please resolve conflicts and re-upload

jacyanthis:
I updated the branch with main, but it will likely become outdated again soon.

Manual testing per template:

Set up an experiment with (i) a participant with a variable (e.g., `treatment` of `A` and `B`) and (ii) a survey stage where a question has a condition based on that variable (e.g., appears if `treatment` equals `A`).

- 1 human participant
- No agent mediators or agent participants required
- Expected behavior: The participant should see the question if and only if their assignment aligns with the condition.

jimbojw (changes requested):
No comments on the code proper, but I'd like to see some unit tests. At least for the `utils/` and `functions/` code. Unit tests for `frontend/` are probably not necessary since it's all components (not standalone utility code).

jacyanthis:
Tests added, thanks!

PR #1168, opened 2026-06-25
fix: private-chat agent prompts are PRIVATE_CHAT type, not CHAT

jacyanthis:
Small fix to make private chat prompt type `StageKind.PRIVATE_CHAT` instead of `StageKind.CHAT` to ensure private chat prompts can be accessed separately.


jacyanthis:
Possible review: @jimbojw
CC: @kenburke @cjqian

jimbojw (approved):

PR #1170, opened 2026-07-09
fix: separate variables are randomized separately

jacyanthis:
Fixes an issue where variables and other randomized objects had dependent, rather than independent, randomization: for example, say that you assign participant variable `A` with values `1`, `2`, `3` and participant variable `B` with values `4`, `5`, `6`...

**Before**: You would only assign participantas`A1+B4`, `A2+B5`, and `A3+B6`.

**After**: You get the full randomization: `A1+B4`, `A1+B5`, `A1+B6`, `A2+B4`, `A2+B5`, etc.…

Everything stays deterministic, and existing functionality is unchanged: the first variable keeps the same sequence it had.

### DIFF
--- a/functions/src/chat/chat.agent.ts
+++ b/functions/src/chat/chat.agent.ts
@@ -614,7 +614,7 @@ async function getNextTurnBasedSpeakerAfterSkippedAgent(
           missingMediators.length > 1
             ? shuffleWithSeed(
                 missingMediators,
-                `${cohortId}-new-mediators-${cycleIndex}`,
+                `${cohortId}-${stage.id}-new-mediators-${cycleIndex}`,
               )
             : missingMediators;
         const nextMediators = [
@@ -626,7 +626,7 @@ async function getNextTurnBasedSpeakerAfterSkippedAgent(
         if (hadMediators) {
           const shuffledParticipants = shuffleWithSeed(
             participantIds,
-            `${cohortId}-${cycleIndex}`,
+            `${cohortId}-${stage.id}-${cycleIndex}`,
           );
           nextTurnOrder = [...nextMediators, ...shuffledParticipants];
         } else {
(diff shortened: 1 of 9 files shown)


jacyanthis:
Possible review: @jimbojw
CC: @kenburke @cjqian

jimbojw (approved):

jimbojw:
## Manual Testing — Local Verification

Checked out the PR locally in a worktree and ran against the Firebase emulators.

### Build & automated tests

All clean:

| Step | Result |
|------|--------|
| `npm ci` | ✅ |
| `npm run build -w utils` | ✅ |
| `npm run build -w functions` | ✅ |
| `npm test -w utils` | ✅ 13 suites, 211 tests |
| `npm test -w frontend` | ✅ 3 suites, 26 tests |
| `npm run test:unit -w functions` | ✅ 7 suites, 134 tests |

### Manual test setup

Created an experiment with:

- **Two random permutation numeric variables** (unexpanded / array mode):
  - `variable_a` — values `[1, 2, 3]`
  - `variable_b` — values `[4, 5]`
  - Both using the default **Cohort** seed strategy, shuffle enabled
- **Three stages**: two chat stages and a survey stage
- The survey stage included a question displaying both resolved variables:
  ```
  Variable A: {{variable_a}}
  Variable B: {{variable_b}}
  ```

### Results

Created four cohorts, each with participants who submitted through both stages.

| Cohort | Participants | Variable A | Variable B |
|--------|-------------|------------|------------|
| 1 | 3 | `3,1,2` | `5,4` |
| 2 | 3 | `3,1,2` | `5,4` |
| 3 | 1 | `2,1,3` | `4,5` |
| 4 | 3 | `1,3,2` | `4,5` |

### Assertions

1. **Fix confirmed** — Within each cohort, Variable A and Variable B are shuffled independently. Before this fix, both variables sharing the same seed strategy would have produced correlated permutations.
2. **Cohort-scoped seed working** — All participants within the same cohort see the same values (expected with the Cohort seed strategy). Different cohorts produce different shuffles.
3. **Cohort 1 = Cohort 2 collision** — With only 6 possible permutations for 3 values and 2 for 2 values, a collision between two cohorts (~8% probability) is expected by chance and not a bug.
4. **Backward compatibility** — The first shuffled variable keeps the bare seed (no namespace suffix), so existing experiments with a single shuffled variable will produce the same sequences they had before.

PR #1171, opened 2026-07-09
feat: observer participants, representative agents, and mediator swapping

jacyanthis:
# Description

Experimenters can assign human participants to observe a Group Chat instead of participating directly. This can include a representative who discusses on behalf of the human.

This is the largest feature branch for our Pluralistic Agents study. To avoid expanding the menus of standard DL functionality, this is done through magic variables that begin with underscores (e.g., `_isObserver`), and users are warned if they coincidentally use these variable names. Some of this branch could be done with templates if that is better, but variables make sense because these would often not be universal across participants; in Pluralistic Agents, we will probably have human participants change between standard participation and observation midwayi

- `_isObserver` (boolean) makes the participant observe group chats instead of being a participant.
- `_hasRepresentative` (boolean) spawns an AI agent representative of the observer in the group chat.
- `_skipPrivateChats` (boolean) makes the participant skip private chats. In Pluralistic Agents, this is necessary for participants to be in a non-observer condition where they have no representative who private-chat interviews them before the group chat.
- `_swapMediator` (string) is the mediator who is swapped in instead of the default mediator.
- `_numOtherAgents` (integer) spawns independent AI agents to participate in the group chat.
- `_numInactivePersonas` (integer) spawns personas who will not actively participate in group chat. In the Pluralistic Agents project, we use this so that a mediator can act as if it had private-chat interviewed other agents. The mediator can refer to them, but they do not actually show up as "participants" (e.g., in the participant list next to the group chat area).

# Additional content

- **Persona bank**: This sets up a persona bank option for these spawned agents to avoid redundant API calls. If a persona bank is present in the experiment, then the personas will be drawn from this bank instead of fresh API calls. If none is present, nothing changes.
- **Turn-based private chats**: This expands the turn-based group chat functionality (#1119) to private chats. In Pluralistic Agents, we use this for interviews, such as an AI agent representative interviewing the human before a group chat. A private chat can now be run by the participant's representative instead of only the default mediator. Because spawning agents takes time, there is also a "setting up" banner displayed so the participant knows it is loading before the discussion itself.
- **Max/min number of messages**: This expands the ability to set maximum and minimum messages in a chat, allowing for more structured observations and interactions.
- **Additional instructions for agent participants**: This also creates an optional `additionalParticipantInstructions` setting on chat stages whose text is appended to AI participants' prompts for that stage, which we use in Pluralistic Agents to encourage persona-based agents to not be overreliant on their personas (e.g., bringing up their occupation in unrelated discussions).
- **Tweaks to turn-based**: This also includes some relatively small quality-of-life tweaks to turn-based functionality, such as defaulting the user’s cursor into the text entry so they don't have to click it each time it's their turn again.

[observers-representatives-dl-260703.webm](https://github.com/user-attachments/assets/72fd8b95-7812-4930-b96b-b2373c741516)


jacyanthis:
Possible review: @jimbojw
CC: @kenburke @cjqian

jimbojw:
@jacyanthis Please resolve conflicts and re-upload

jacyanthis:
> @jacyanthis Please resolve conflicts and re-upload

@jimbojw I don't see any conflicts, but I just synced the branch in case that's what you were referring to. Anything else I can do to move this forward?

jacyanthis:
Testing general functionality with manual testing per template:

Set up an experiment with (i) a participant who is assigned the variables `{"_isObserver": true, "_hasRepresentative": true, "_numOtherAgents": 2}`, (ii) a transfer stage with its treatment index set to `0`, and (iii) a turn-based group chat stage with an agent mediator that follows the transfer stage.

Expected behavior: When the participant reaches the chat stage, the agents should be spawned, and the participant should observe rather than directly participate in the chat. The participant's text entry should be disabled, and a representative agent should speak on their behalf with two other agents participating.

PR #1174, opened 2026-07-13
feat: the participant is quizzed periodically during a group chat

jacyanthis:
This PR builds on #1171 and should only be merged after it is merged.

This PR adds a periodic quiz for the participant to share their reactions to the group chat. By default, this pauses the turn-based group chat after each third of the minimum message count (up to 3 quizzes; fewer if the minimum is fewer than 3). It asks two questions: a 7-point Likert ("Do you like the process so far?") and a free-text explanation ("In one sentence, describe what you currently like or dislike.").

This is gated by a magic participant variable, `_isQuizzed` (boolean). If the chat is not turn-based, then this variable is ignored, and the variable editor warns the experimenter that it is being ignored.

quiz-example

### DIFF
--- a/docs/assets/api/schemas.json
+++ b/docs/assets/api/schemas.json
@@ -741,6 +741,14 @@
         },
         "additionalParticipantInstructions": {
           "type": "string"
+        },
+        "minNumberOfMessages": {
+          "minimum": 0,
+          "type": "integer"
+        },
+        "maxNumberOfMessages": {
+          "minimum": 1,
+          "type": ["integer", "null"]
         }
       }
     },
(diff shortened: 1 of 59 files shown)


jacyanthis:
Possible review: @jimbojw
CC: @kenburke @cjqian

jacyanthis:
Manual testing per template:

Set up an experiment with (i) a participant who is assigned the boolean variable `_isQuizzed` as `true` and (ii) a turn-based group chat stage with an agent mediator and a minimum message count of 3. No agent participants are required.

Expected behavior: The quiz pauses the chat after each of the first three messages, resuming each time once the quiz is submitted.

PR #1176, opened 2026-07-13
Add message replies and reactions to group chat

liuziyi219:
Participants can reply to a message with a quote of it, and react to messages with a heart or thumbs up. Reactions record **who** reacted, not just how many.

The feature is **opt-in per group chat stage** and **off by default** — see _Enabling the feature_ below.

## Enabling the feature

`ChatStageConfig.enableReactionsAndReplies` (optional boolean, default `false`) gates the whole feature for a given group chat stage. It is surfaced as a "Reactions and replies" checkbox in the group chat editor's Conversation settings, next to the existing turn-based toggle.

- When **off** (default), participants see plain messages: no react/reply buttons, no reaction chips, and no reply quotes. Gating happens in the participant UI (`chat_message` returns `nothing` for the actions row and the reply quote), so a stage that never opted in behaves exactly as it did before this PR.
- When **on**, the full react/reply UI described below is shown.

The flag lives on the stage config so each group chat stage can enable it independently, and so the setting travels with the experiment definition and templates. Because it is a new field on an exported config type, `schemas.json` and the generated Python `types.py` are regenerated to match.

Gating is currently front-end only; the write endpoints are unchanged. A follow-up could enforce the flag server-side (reject reactions / strip `replyTo` when a stage has not opted in) so the toggle is authoritative rather than cosmetic.

## Data model

Two optional fields on `ChatMessage`, so existing chat documents remain valid:

- `reactionMap: {heart: string[], thumbsUp: string[]}` — each reaction maps to the public IDs of everyone who applied it. The list of reactors *is* the storage and the count is its length, so a count can never drift out of sync with who reacted.
- `replyTo: {id, senderId, name, message}` — a snapshot of the quoted message rather than just a parent ID, so the quote still renders if the original is not in the loaded history, and it survives into experiment downloads. Quotes are truncated (280 chars) so a chain of replies cannot grow each message without bound.

## Writes

A new `updateChatMessageReaction` callable applies reactions with Firestore's `arrayUnion`/`arrayRemove`. These are atomic, so simultaneous reactions from different participants cannot overwrite each other, and idempotent, so a double-click cannot double-count. Chat documents were already client-unwritable in the security rules and stay that way — no rules change.

## Data export

Replies and reactions flow into the JSON experiment download automatically, since `data.ts` passes the raw documents through.

The chat history **CSV** builds its columns explicitly, so it would have silently dropped both — anyone analysing a conversation from the CSV would have concluded nothing was recorded. Added seven columns: `Reply to message ID`, `Reply to sender ID`, `Reply to content`, `Heart count`, `Heart by`, `Thumbs up count`, `Thumbs up by`. Reactor lists are semicolon-separated because `toCSV` strips commas from cells (a comma would shift every following column).

## Drive-by fix

`CohortService.loadChatMessages` appended every changed document to the message list without checking the change type. Only creates happened before, so it went unnoticed — but a reaction *updates* a chat document, which would have rendered that message twice. The listener now rebuilds from the snapshot, as the private chat listener already does.

## Testing

- 7 integration tests against the Firestore emulator covering the reaction write path, including 8 participants reacting simultaneously (a read-modify-write would drop some; `arrayUnion` keeps all), un-reacting removing only the caller, and reacting to a missing message.
- The deployed callable driven over HTTP end-to-end: toggling, reactions staying independent, plus `NOT_FOUND` on a missing message and `INVALID_ARGUMENT` on an unknown reaction or empty sender.
- Unit tests for the quote/reaction helpers, including truncation and the backward-compatibility case (messages saved before this feature have no `reactionMap` and read as zero reactions).
- CSV tests pinning the header row against a real message row so the columns cannot drift out of alignment.
- Verified in the running app with two participants, and confirmed in a real experiment download that reactions and replies appear in both the JSON and the CSV.
- Verified the opt-in toggle in the running app: with the stage's "Reactions and replies" setting off (the default), no react/reply affordances appear; with it on, the full feature is shown.

Full suites pass on this base: 107 emulator, 219 utils, 134 functions, 30 frontend.

## Note for reviewers

The reaction endpoint takes `senderId` from the client, matching the existing convention in `createChatMessage`. This does not weaken the current trust model, but if reactions are intended as attributable research data, deriving the sender server-side would be the thing to tighten.

### DIFF
--- a/docs/assets/api/schemas.json
+++ b/docs/assets/api/schemas.json
@@ -735,6 +735,9 @@
         },
         "isTurnBased": {
           "type": "boolean"
+        },
+        "enableReactionsAndReplies": {
+          "type": "boolean"
         }
       }
     },
--- a/frontend/src/components/chat/chat_input.scss
+++ b/frontend/src/components/chat/chat_input.scss
@@ -6,6 +6,12 @@
   width: 100%;
 }
 
+.chat-input {
+  @include common.flex-column;
+  gap: common.$spacing-small;
+  width: 100%;
+}
+
 .input {
   align-items: center;
   background: var(--md-sys-color-surface);
@@ -38,4 +44,40 @@
       pointer-events: none;
     }
   }
+}
+
+.reply-banner {
+  align-items: center;
+  background: var(--md-sys-color-surface);
+  border: 1px solid var(--md-sys-color-outline-variant);
+  border-radius: common.$spacing-medium;
+  box-sizing: border-box;
+  display: flex;
+  gap: common.$spacing-small;
+  justify-content: space-between;
+  padding: common.$spacing-small common.$spacing-small common.$spacing-small
+    common.$spacing-large;
+  width: 100%;
+}
+
+.reply-preview {
+  @include common.flex-column;
+  border-left: 3px solid var(--md-sys-color-outline-variant);
+  gap: 2px;
+  min-width: 0;
+  padding-left: common.$spacing-medium;
+}
+
+.reply-author {
+  @include typescale.label-small;
+  color: var(--md-sys-color-on-surface-variant);
+  font-weight: 500;
+}
+
+.reply-text {
+  @include typescale.label-small;
+  color: var(--md-sys-color-outline);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
--- a/frontend/src/components/chat/chat_input.ts
+++ b/frontend/src/components/chat/chat_input.ts
@@ -1,6 +1,6 @@
 import {MobxLitElement} from '@adobe/lit-mobx';
 
-import {CSSResultGroup, html} from 'lit';
+import {CSSResultGroup, html, nothing} from 'lit';
 import {customElement, property, state} from 'lit/decorators.js';
 
 import {core} from '../../core/core';
@@ -26,8 +26,12 @@ export class ChatInputComponent extends MobxLitElement {
     input: string,
   ) => {
     if (!this.stageId || !input || input.trim() === '') return;
-    await this.participantService.createChatMessage({message: input.trim()});
+    await this.participantService.createChatMessage({
+      message: input.trim(),
+      replyTo: this.participantAnswerService.getChatReply(this.stageId),
+    });
     this.participantAnswerService.updateChatInput(this.stageId, '');
+    this.participantAnswerService.clearChatReply(this.stageId);
   };
   @property() storeUserInput: (input: string) => void = (input: string) => {
     if (!this.stageId) return;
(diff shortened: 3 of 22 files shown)


jimbojw:
Thanks for including tests. What are the steps to manually test?

Please include:

* Required and optional Stages that the Experiment should have.
* Minimum number of Human Participants needed in a Cohort.
* Whether a Mediator is required and if so, any special configuration?
* What each Participant should do.
* What is the expected behavior of the system to indicate the change was successful?

liuziyi219:
Thanks for the review — here is the setup and expected behavior for reproducing this.

### Stages the experiment should have
- **Profile stage** (required) — participants need a name and avatar, which the feature uses for the reply author label and reaction attribution. Either `ANONYMOUS_ANIMAL` or `DEFAULT` profile type works.
- **Chat stage** (required) — this is the group chat where replies and reactions live.

This is exactly the built-in **"Group Chat Experiment"** quickstart template (`getQuickstartGroupChatTemplate`), so no custom stage setup is needed. Any additional stages are irrelevant to the feature.

### Minimum number of human participants per cohort
- **2** to exercise the feature as intended — reacting to and replying to another participant's message, and confirming it appears live in the other participant's window.
- 1 is enough to see the UI render and to verify persistence, since a participant can react/reply to any message including their own (there is no self-exclusion).

### Mediator
**Not required.** The feature is independent of mediation. It can be tested with the plain "Group Chat Experiment" template (no agents). It also works unchanged with the mediated template ("Mediated Group Chat Experiment") — a mediator's messages can be reacted to and replied to like any other, no special configuration.

### What each participant does
1. Complete the profile stage, enter the group chat, and send a message or two.
2. Hover a message — a reply arrow and the ❤️ / 👍 reaction buttons appear.
3. Click a reaction to toggle it on; click again to remove it.
4. Click reply — a quote banner appears above the input; type a response and send.

### Expected behavior indicating success
- **Reaction:** the chip shows a count and highlights as active; its tooltip names who reacted ("You", others by name). It appears **live** in the other participant's window, and the reacted-to message is **not** duplicated in the history (this was a latent listener bug the PR also fixes).
- **Reply:** the sent message renders with the quoted text above its bubble, in both windows.
- **Persistence:** in the Firestore emulator, the chat doc under `.../publicStageData/{stageId}/chats/{chatId}` gains `reactionMap: {heart: [...publicIds], thumbsUp: [...]}` and, for replies, a `replyTo` object.
- **Export:** downloading the experiment shows reactions and replies in both `{ExperimentName}.json` (under `cohortMap → chatMap`) and the `_ChatHistory_*.csv` (new `Heart by` / `Thumbs up by` / `Reply to content` columns).

jimbojw:
@liuziyi219 OK thanks, I was able to run it locally and perform a manual test according to your instructions.

One thing I notice: It's possible to react to one's own message. Is that intentional?

jimbojw (approved):

